Issue #36, opened 2012-03-03
provide a headless rtorrentd


infinity0:
Would it be possible to provide a daemon version of rtorrent that doesn't have a UI and just sits there waiting for XML RPC requests? At the moment, I need to mess around with screen - if i just do rtorrent & disown, XML RPCs hang for some reason.

It doesn't even need to background itself, since start-stop-daemon can handle this pretty easily. (Actually, if you did make it background itself by using fork(), you would also need to create a pidfile somewhere so start-stop-daemon can get at the new pid).

It would be really useful for scripting things with torrents. transmission-daemon is the only other client that comes close in this regard, but their RPC client doesn't output machine-parseable data, and for some reason it's not connecting properly for my setup (but rtorrent does). Also they don't support the http_cacert options which is pretty important for what I want to do (encrypted backup over bittorrent).

awh-tokyo:
Is there any reason this pull request was never accepted?  What I'm doing now (with screen) just seems clunky.

rolfyone:
I'm pretty keen on this enhancement also... I might fork and import it myself...

munnerz:
Looks like a PR has been merged for this, just awaiting it landing in a release: https://github.com/rakshasa/rtorrent/pull/446

rraallvv:
Is headless mode supported in 0.9.6 ?

chros73:
No, only in 0.9.7.

rraallvv:
@chros73 Thanks. Nevertheless, I was able to use screen with 0.9.6 for a "sort of" headless mode.

chros73:
Yes, or tmux. That's how the majority of people use rtorrent.

rraallvv:
@chros73 Awesome! I'll check out tmux too then. Thanks.

phxn:
tmux gets my vote too, it’s extremely useful.

PR #111, opened 2013-01-15
Allow rtorrent to run as a daemon without screen.  implements issue #36

sallyswiss:
this checks if running in a terminal before initializing ncurses with initscr and then does not initialize if not a terminal.  this means rtorrent can run as a daemon but will still function normally under normal conditions.

### DIFF
--- a/src/control.cc
+++ b/src/control.cc
@@ -115,7 +115,9 @@ Control::initialize() {
 
   m_ui->init(this);
 
-  m_inputStdin->insert(torrent::main_thread()->poll());
+  if(display::Canvas::have_term()) {
+    m_inputStdin->insert(torrent::main_thread()->poll());
+  }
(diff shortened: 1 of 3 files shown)


rakshasa:
In any case the 'have term' needs to be a variable that can be set (see src/command_*.cc) and default to true.

wheelerlaw:
What is the status of this?

rakshasa:
Status is that the patch defaults to daemon while that should be the non-default option set by a properly named system.daemon.enable variable (or something).

For someone who what's this merged, forged the branch, merge the latest develop, then add the (very easy to add) variable and checks to have the proper behavior.

Make a new pull request, attribute the original developers above, and make something I want to merge.

o6uoq:
+1

PR #280, opened 2015-03-11
Partially update outdated documentation

mynamewastaken:


pyroscope:
When https://github.com/rakshasa/rtorrent/wiki/CONFIG-Template is ready, `doc/rtorrent.rc` should simply be removed (not copied/updated, so that the master copy is still in the Wiki and editable).

PR #291, opened 2015-05-04
Add load.raw_start_verbose

### LINKED ISSUE #289: Missing load method with raw+start+verbose flags
Current options found here: https://github.com/rakshasa/rtorrent/blob/7537a3c2a91d0915f1c4c89b01cd583629dc5fd4/src/command_events.cc#L347-L355

Would propose adding `load.raw_start_verbose`.

RSully:
Fixes #289.


rakshasa:
It's missing 'core::Manager::create_quiet' and please add the line break same as in the line above..

RSully:
I thought the difference between verbose or not was that quiet flag. Otherwise, I'm at a loss as to what verbose actually does. Originally I thought verbose is what allowed you to pass additional commands (e.g. `d.directory.set=`) but that doesn't seem to be the case. Thanks.

rakshasa:
Sorry, brainfart.

PR #405, opened 2016-03-08
added ui.current_view command

pyroscope:
Allow querying the selected view, to change+restore it, or rotate through a set of views.


rakshasa:
As long as the code looks good and I can trust it to compile I'll keep merging little things like this. ;)

PR #420, opened 2016-04-09
Apply color patch from rtorrent-0.9.6_color.patch

myallod:
Please add color to rtorrent

### DIFF
--- a/src/command_network.cc
+++ b/src/command_network.cc
@@ -237,6 +237,25 @@ initialize_command_network() {
   core::CurlStack* httpStack = control->core()->http_stack();
 
   CMD2_ANY_STRING  ("encoding.add", std::bind(&apply_encoding_list, std::placeholders::_2));
+  CMD2_ANY         ("colors.color_inactive_fg",       std::bind(&ui::Root::get_color_inactive_fg,   control->ui()));
+  CMD2_ANY_VALUE_V ("colors.color_inactive_fg.set",   std::bind(&ui::Root::set_color_inactive_fg,   control->ui(), std::placeholders::_2));
+  CMD2_ANY         ("colors.color_inactive_bg",       std::bind(&ui::Root::get_color_inactive_bg,   control->ui()));
+  CMD2_ANY_VALUE_V ("colors.color_inactive_bg.set",   std::bind(&ui::Root::set_color_inactive_bg,   control->ui(), std::placeholders::_2));
+
+  CMD2_ANY         ("colors.color_dead_fg",       std::bind(&ui::Root::get_color_dead_fg,   control->ui()));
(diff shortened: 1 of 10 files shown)


pyroscope:
Even if Jari accepts this, doing it as an array with names or enum indices is way less code and more extensible.

rakshasa:
As said, array and enums is a must.

It also needs to have a proper setting to enable/disable colors, and not to mention this should obviously be in command_ui.cc, under the 'ui.foo.bar' command namespace.

I'm all for adding colors if the code is good.

myallod:
would you please give me a hint about array and enum that needed.

chros73:
You can take a look pyroscope's beautiful code in [rtorrent-ps](https://github.com/pyroscope/rtorrent-ps) : https://github.com/pyroscope/rtorrent-ps/blob/master/patches/ui_pyroscope.cc

KopfKrieg:
Just for your information: The "official" colorpatch branch is avaliable in my fork of the repository (including prior versions, that's also the one used in the Arch Linux rtorrent-color AUR package): https://github.com/KopfKrieg/rtorrent/tree/colorpatch

KopfKrieg:
Since my C++ skills are very basic and I have trouble porting the v0.9.6 colorpatch to the shortly released v0.9.7: Is there somebody who could help me rewrite the colorpatch so it could officially be added to rtorrent? Including the above mentioned changes to add the code to the command_ui.cc and using enums (and adding an option to disable colors, which should probably be the default?)?

pyroscope:
Look at `torrent.list.layout` and do something similar for colors – a bool would suffice, but a string value would allow different colorization schemes / themes (possibly foregoing variables for different single colors).

Or take my "palette" of functional colors (or parts of it):

* https://github.com/pyroscope/rtorrent-ps/blob/f2433b3aef285c650372c0c9804068361b3e7577/patches/ui_pyroscope.h#L24-L56
* https://github.com/pyroscope/rtorrent-ps/blob/f2433b3aef285c650372c0c9804068361b3e7577/patches/ui_pyroscope.h#L24-L56
* https://github.com/pyroscope/rtorrent-ps/blob/42989583ec54a3ad6c81ef83c92a52495d92a395/patches/ui_pyroscope.cc#L1103-L1136

… and code scatteredc around the module using those colors.

Hund:
What's the progress of this?

KopfKrieg:
> What's the progress of this?

Well, I don't maintain it any more, and it doesn't seem like anyone has tried to port it to v0.9.7 yet. So … still the same as 3 years ago, probably.

rakshasa:
A color patch has been merged.

Hund:
Better late than never. :D Thank you!

PR #447, opened 2016-06-23
Display values of 1 throttle.up in the first part of status bar

chros73:
@rakshasa If you don't like this idea you can reject it.
It's useful especially if you modify its value dynamically, like with this:
https://github.com/chros73/rtorrent-ps_setup/blob/master/ubuntu-14.04/home/chros73/.rtorrent.rc#L221
https://github.com/chros73/rtorrent-ps_setup/blob/master/ubuntu-14.04/home/chros73/bin/getLimits.sh

----------------
Display values of 1 throttle.up in the first part of status bar.
Currently the beginning of status bar looked like this:
`[Throttle 500/1500 KB]  [Rate: 441.6/981.3 KB]`
Include the max limit of the throttle, the main upload rate and the upload rate of the throttle (in this order):
`[Throttle 500 (200)/1500 KB]  [Rate: 441.6 (190.0|51.6)/981.3 KB]`
- make the displayable throttle.up name configurable via config file: `ui.status.throttle_up_name.set`
- don't display this extra info in the following cases:

-- there isn't any throttle.up name as the config variable suggest or the given name is "NULL"
-- the throttle.up is not throttled (=0)
-- the global upload is not throttled (=0) (the throttle.up won't be taken into account in this case)

### DIFF
--- a/src/command_ui.cc
+++ b/src/command_ui.cc
@@ -584,4 +584,6 @@ initialize_command_ui() {
 
   CMD2_ANY_LIST ("elapsed.less",         std::bind(&apply_elapsed_less, std::placeholders::_2));
   CMD2_ANY_LIST ("elapsed.greater",      std::bind(&apply_elapsed_greater, std::placeholders::_2));
+
+  CMD2_VAR_STRING("ui.status.throttle_up_name", "");
 }
--- a/src/display/utils.cc
+++ b/src/display/utils.cc
@@ -257,21 +257,62 @@ print_client_version(char* first, char* last, const torrent::ClientInfo& clientI
   }
 }
 
+int64_t
+retrieve_throttle_up_value(const torrent::Object::string_type& name, bool rate) {
+  core::ThrottleMap::iterator itr = control->core()->throttles().find(name);
+
+  if (itr == control->core()->throttles().end()) {
+    return (int64_t)-1;
+  } else {
+    torrent::Throttle* throttle = itr->second.first;
+    int64_t throttle_max = (int64_t)throttle->max_rate();
+    if (rate) {
+      if (throttle_max > 0)
+	return (int64_t)throttle->rate()->rate();
+      else
+	return (int64_t)-1;
+    } else {
+      return throttle_max;
+    }
+  }
+}
+
 char*
 print_status_info(char* first, char* last) {
-  if (!torrent::up_throttle_global()->is_throttled())
+  std::string throttle_up_name = rpc::call_command_string("ui.status.throttle_up_name").c_str();
+
+  if (!torrent::up_throttle_global()->is_throttled()) {
     first = print_buffer(first, last, "[Throttle off");
-  else
+  } else {
     first = print_buffer(first, last, "[Throttle %3i", torrent::up_throttle_global()->max_rate() / 1024);
+    if (!throttle_up_name.empty() && throttle_up_name != "NULL") {
+      int64_t throttle_up_max = retrieve_throttle_up_value(throttle_up_name, false);
+      if (throttle_up_max > 0)
+        first = print_buffer(first, last, " (%1.0f)", (double)throttle_up_max / 1024.0);
+    }
+  }
 
   if (!torrent::down_throttle_global()->is_throttled())
     first = print_buffer(first, last, "/off KB]");
   else
     first = print_buffer(first, last, "/%3i KB]", torrent::down_throttle_global()->max_rate() / 1024);
-  
-  first = print_buffer(first, last, " [Rate %5.1f/%5.1f KB]",
-                       (double)torrent::up_rate()->rate() / 1024.0,
-                       (double)torrent::down_rate()->rate() / 1024.0);
+
+  double global_uprate = (double)torrent::up_rate()->rate() / 1024.0;
+  first = print_buffer(first, last, " [Rate %5.1f", global_uprate);
+
+  if (!throttle_up_name.empty() && throttle_up_name != "NULL" && torrent::up_throttle_global()->is_throttled()) {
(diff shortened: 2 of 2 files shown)


chros73:
This is how it look like (using rtorrent-ps). Note: using spaces are intentional (otherwise it was too much info).
rtorrent-ps-ch-0 9 6-happy-pastel-kitty-s

rakshasa:
Looks good, just a couple of things to fix; change it to add the space on each side of the slash instead (including for the default display), and some minor code changes.

src/command_ui.cc
```diff
@@ -584,4 +584,6 @@ initialize_command_ui() {
 
   CMD2_ANY_LIST ("elapsed.less",         std::bind(&apply_elapsed_less, std::placeholders::_2));
   CMD2_ANY_LIST ("elapsed.greater",      std::bind(&apply_elapsed_greater, std::placeholders::_2));
+
+  CMD2_VAR_STRING("ui.status.throttle_up_name", "");
```
rakshasa: This should be 'ui.status.throttle.up.name', and should also include the option for the down throttle.

src/display/utils.cc
```diff
@@ -257,21 +257,62 @@ print_client_version(char* first, char* last, const torrent::ClientInfo& clientI
...
+    } else {
+      return throttle_max;
+    }
+  }
+}
+
 char*
 print_status_info(char* first, char* last) {
-  if (!torrent::up_throttle_global()->is_throttled())
+  std::string throttle_up_name = rpc::call_command_string("ui.status.throttle_up_name").c_str();
```
rakshasa: Move this inside the 'else' block so you only initialize the variable right before you use it, this isn't C.

Also, no need to use 'c_str' conversion here, just use the std::string it returns.

src/display/utils.cc
```diff
@@ -257,21 +257,62 @@ print_client_version(char* first, char* last, const torrent::ClientInfo& clientI
...
+
+  if (!torrent::up_throttle_global()->is_throttled()) {
     first = print_buffer(first, last, "[Throttle off");
-  else
+  } else {
     first = print_buffer(first, last, "[Throttle %3i", torrent::up_throttle_global()->max_rate() / 1024);
+    if (!throttle_up_name.empty() && throttle_up_name != "NULL") {
```
rakshasa: There's no need for any special "NULL" name handling here, just use 'empty' as meaning that we don't display any additional information.

src/display/utils.cc
```diff
@@ -257,21 +257,62 @@ print_client_version(char* first, char* last, const torrent::ClientInfo& clientI
...
     first = print_buffer(first, last, "/%3i KB]", torrent::down_throttle_global()->max_rate() / 1024);
-  
-  first = print_buffer(first, last, " [Rate %5.1f/%5.1f KB]",
-                       (double)torrent::up_rate()->rate() / 1024.0,
-                       (double)torrent::down_rate()->rate() / 1024.0);
+
+  double global_uprate = (double)torrent::up_rate()->rate() / 1024.0;
+  first = print_buffer(first, last, " [Rate %5.1f", global_uprate);
+
+  if (!throttle_up_name.empty() && throttle_up_name != "NULL" && torrent::up_throttle_global()->is_throttled()) {
```
rakshasa: Same with NULL here.

src/display/utils.cc
```diff
@@ -257,21 +257,62 @@ print_client_version(char* first, char* last, const torrent::ClientInfo& clientI
...
+    int64_t throttle_uprate_b = retrieve_throttle_up_value(throttle_up_name, true);
```
rakshasa: The '_b' postfix doesn't tell us anything about what is special about that throttle variable, be more descriptive.

Also please add some empty lines between variables and 'if' statements, it makes it easier to read.

chros73:
> change it to add the space on each side of the slash instead

A. Do you mean this?
`[Throttle 500(200) / 1500 KB] [Rate: 441.6(190.0|51.6) / 981.3 KB]`

B. Or this?
`[Throttle 500 (200) / 1500 KB] [Rate: 441.6 (190.0|51.6) / 981.3 KB]`

> There's no need for any special "NULL" name handling here ... Same with NULL here.

Actually there is. It was a business decision from my part to not show values in this case, therefore there isn't any check for this in: `retrieve_throttle_up_value` .

> and some minor code changes ... and should also include the option for the down throttle

Unfortunately, these changes won't be minor, so I can't do them in the foreseeable future (I'm lack of spare time).

If you like I can do just the really minor changes and later the rest can be done by somebody else.

rakshasa:
I was thinking of A as that makes it more obvious what is up, and what is down.

Not sure what you point about the NULL edge case is, my assumption is that if the value is an empty string then it means you don't show anything beyond the default values, e.g. doesn't involve any throttles except the global one.

I don't mind if you don't add the down throttle part, I can copy-paste the code well enough as long as the up throttle is implemented properly.

chros73:
OK, I'll look into it when I'll have spare time. Cheers

chros73:
> Not sure what you point about the NULL edge case is, my assumption is that if the value is an empty string

But here `"NULL"` is string not the `NULL` pointer.

From the [old trac wiki](http://web.archive.org/web/20131104120054/http://libtorrent.rakshasa.no/wiki/RTorrentCommonTasks#Usingcustomthrottles) : "The NULL throttle is a special unlimited throttle that bypasses the global throttle."

So in this case the global values has to be taken care of, and I don't know how it works, that's why (along with the [use case](https://github.com/rakshasa/rtorrent/wiki/Favoring-group-of-torrents)) I decided not to show them in this case.

chros73:
> Not sure what you point about the NULL edge case is ...

You were right about it, sorry. :)

I'm closing this pull request, since I made all the required changes, and creating a new one.

PR #454, opened 2016-07-09
Travis config fix

Speeddymon:
Travis config seems broken. Try to fix it.

### DIFF
--- a/src/core/download_list.cc
+++ b/src/core/download_list.cc
@@ -607,6 +607,7 @@ DownloadList::received_finished(Download* download) {
     // Set some 'checking_finished_thingie' variable to make hash_done
     // trigger correctly, also so it can bork on missing data.
     hash_queue(download, Download::variable_hashing_last);
+    confirm_finished(download);
   else
     confirm_finished(download);
 }
--- a/src/core/download_list.cc
+++ b/src/core/download_list.cc
@@ -606,8 +606,8 @@ DownloadList::received_finished(Download* download) {
   if (rpc::call_command_value("pieces.hash.on_completion"))
     // Set some 'checking_finished_thingie' variable to make hash_done
     // trigger correctly, also so it can bork on missing data.
-    hash_queue(download, Download::variable_hashing_last);
-    confirm_finished(download);
+    if (hash_queue(download, Download::variable_hashing_last))
+      confirm_finished(download);
   else
     confirm_finished(download);
 }
--- a/src/core/download_list.cc
+++ b/src/core/download_list.cc
@@ -606,8 +606,7 @@ DownloadList::received_finished(Download* download) {
   if (rpc::call_command_value("pieces.hash.on_completion"))
     // Set some 'checking_finished_thingie' variable to make hash_done
     // trigger correctly, also so it can bork on missing data.
-    if (hash_queue(download, Download::variable_hashing_last))
-      confirm_finished(download);
+    hash_queue(download, Download::variable_hashing_last);
   else
     confirm_finished(download);
 }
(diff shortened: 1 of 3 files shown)


src/core/download_list.cc
```diff
@@ -606,8 +606,8 @@ DownloadList::received_finished(Download* download) {
   if (rpc::call_command_value("pieces.hash.on_completion"))
     // Set some 'checking_finished_thingie' variable to make hash_done
     // trigger correctly, also so it can bork on missing data.
-    hash_queue(download, Download::variable_hashing_last);
-    confirm_finished(download);
+    if (hash_queue(download, Download::variable_hashing_last))
+      confirm_finished(download);
```
pyroscope: This is broken, and one of the reasons you ALWAYS use {}, even on one-liners in then/else.

Speeddymon:
```
  if (rpc::call_command_value("pieces.hash.on_completion"))
    // Set some 'checking_finished_thingie' variable to make hash_done
    // trigger correctly, also so it can bork on missing data.
    if (hash_queue(download, Download::variable_hashing_last))
      confirm_finished(download);
  else
    confirm_finished(download);
```

The above code was exactly how I intended it to be.

With braces:

```
if (rpc::call_command_value("pieces.hash.on_completion"))
{
    // Set some 'checking_finished_thingie' variable to make hash_done
    // trigger correctly, also so it can bork on missing data.
    if (hash_queue(download, Download::variable_hashing_last)) // if this returns true then we confirm finished, else we don't
    { confirm_finished(download); }
}
else // if pieces.hash.on_completion is disabled
{ confirm_finished(download); }

Admittedly, I'm not a programmer, so if this is flawed code, hopefully the intent is conveyed in the comments I have added and someone with better skill can fix it.

rakshasa:
Note that while there are 'deprecated' (redirected/alias) commands still in src/main.cc, those that are not deprecated are considered so common that they have the abbreviated form available.

Others have been deprecated as the syntax and/or arguments used have changed.

At some point I (or someone who wants to do some coding) will add a new config file system that will turn the equivalent of 'foo.bar = 123' or 'foo.bar: 123' into a call to 'foo.bar.set(123)'.

pyroscope:
A real ground-breaking change would be to use Lua for config + scripting, and call the commands from there — rtorrent 1 maybe?

rakshasa:
Lua might be added, however too much code relies on my hacked up commands.

If I can at the very least replace all the 'foo.bar="baz"' in the source with 'foo.bar("baz")' it would be a step forward.

Speeddymon:
If the only obstacle is a lack of time to go through and make the aforementioned change -- 'foo.bar="baz"' to 'foo.bar("baz")' -- I'll be happy to go through and submit a pull request to make those changes. :-)

It's simple enough that I should be able to get it done within a week or so after I get back from holiday.

rakshasa:
My recommendation would be to either use emacs regexp or do it using sed and some scripts. That's how I usually deal with large codebase changes.

Speeddymon:
That's fine. Just want to make sure the syntax is absolutely correct.

What's posted above will apply to 99 or more percent of all of the changes that need to be made, right?

rakshasa:
Actually you need to include the ".set" for both sides.

Which is why it's best to actually save the (bash/regex) scripts, as running them is the easy part. What I want to get rid of is the equal sign used in the code, and the parenthesis is a more robust way of indicating what are the variables passed to that specific command.

rakshasa:
And right, the reason I haven't done it is that if there's corner-cases I'll be having to figure out what commands got messed up.

Speeddymon:
Yup, I included the .set in the example I last posted. ;-)

I'll happily test corner cases (hopefully Travis will catch them) before I submit the pull req. I'll break them up into several pull requests for ease of isolating those changes which break something.

pyroscope:
Since Lua was mentioned here, see #490.

PR #474, opened 2016-08-03
Merzlya display

rakshasa:
Cleaned up merzlya's compact torrent display patch.

### DIFF
--- a/doc/rtorrent.rc
+++ b/doc/rtorrent.rc
@@ -3,59 +3,74 @@
 # uncomment the options you wish to enable.
 
 # Maximum and minimum number of peers to connect to per torrent.
+#
 #throttle.min_peers.normal.set = 40
 #throttle.max_peers.normal.set = 100
 
 # Same as above but for seeding completed torrents (-1 = same as downloading)
+#
 #throttle.min_peers.seed.set = 10
 #throttle.max_peers.seed.set = 50
 
 # Maximum number of simultanious uploads per torrent.
+#
 #throttle.max_uploads.set = 15
 
 # Global upload and download rate in KiB. "0" for unlimited.
+#
 #throttle.global_down.max_rate.set_kb = 0
 #throttle.global_up.max_rate.set_kb = 0
 
 # Default directory to save the downloaded torrents.
+#
 #directory.default.set = ./
 
 # Default session directory. Make sure you don't run multiple instance
 # of rtorrent using the same session directory. Perhaps using a
 # relative path?
+#
 #session.path.set = ./session
 
 # Watch a directory for new torrents, and stop those that have been
 # deleted.
+#
 #schedule2 = watch_directory,5,5,load.start=./watch/*.torrent
 #schedule2 = untied_directory,5,5,stop_untied=
 
 # Close torrents when diskspace is low.
+#
 #schedule2 = low_diskspace,5,60,close_low_diskspace=100M
 
 # The ip address reported to the tracker.
+#
 #network.local_address.set = 127.0.0.1
 #network.local_address.set = rakshasa.no
 
 # The ip address the listening socket and outgoing connections is
 # bound to.
+#
 #network.bind_address.set = 127.0.0.1
 #network.bind_address.set = rakshasa.no
 
 # Port range to use for listening.
+#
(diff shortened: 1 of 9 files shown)


mvucBmM0:
The "L" keybind is crashing with:
"rtorrent: Command "download.list.layout" does not exist."

rakshasa:
If "L" doesn't work, it should be a pretty easy pull request to create. I'm starting to think the way to get back into coding on rtorrent and libtorrent is to not touch those things as that forces people to do things to fix their favorite pet peeve.

mvucBmM0:
It's not a pet peeve, i was just reporting the error.
I'm sorry if you're not in a good day...

rakshasa:
No, what I'm saying is that it's easy to create a pull request fixing that minor issue. Which is something people who are affected by an issue that is easy to fix should really do as a favor to the people (or person in this case) that need to manage a project... in this case, a project far too big for just one individual.

Basically fixing the command tied to the L key should be so easy that I should have a pull request waiting for me, not a ticket telling me I need to fix some code.

If you think this way of approaching the project is a bit dickish then you should see my inbox. I'm trying to get more people to not just create issues for small and easily fixed issues, but to actually make them give me an easy merge request that I can just click a couple of buttons and get done with.

Put in some effort and it will be part of rtorrent/libtorrent. Don't, and it's probably going to be ignored. When it came to this particular pull request I'd already spent about an hour or so refactoring the code.

It isn't my intention to be 'in a bad mood', it is rather that having put in the kind of effort that I've done, I'm not really that forgiving of people who should know how to fix a minor detail not put in the small amount of work needed to create something I can easily merge.

One road leads to burnout, the other to a more active community. If I need to sound peeved to make the latter happen then so be it.

mvucBmM0:
I understand what you're saying.
Unfortunately i don't know C, but i'll see if i can help.
Sorry if i was annoying in any way. I appreciate your work. A lot.

chros73:
I understand what you mean and that should be the power of an opensource project, especially with help of GitHub, but I don't think this rushing world works like this, especially nowdays. :(

And broken/untested code should never get into master branch :) (I don't want you be pissed off, more than that, cheer you up! Thanks)

rakshasa:
The thing is that the above issue is just a string replace in one part of the code that should be easy to fix.

If I have to do everything then I'm more likely to do nothing. And unless there's CI for every part of rtorrent I'm relying on the users to point out / fix issues. More of the latter is better, as I'm less likely to continue working on this if I have to fix all of the former.

pyroscope:
Related to that, people who cannot code C++ can still help with docs (especially the wiki). By doing so, you free up time of people who CAN write C++, thus helping yourself also (by getting new features). If the current activity level continues, or even increases, we could end up with new, documented features. :smirk_cat:

rakshasa:
Well, in this case you don't even need to know C++ nor C, the issue is just a string that needs replacing.

As long as you know how GitHub works the task is not that different from doing documentation. For those who might question why I haven't done it, while instead writing this much in tickets, it is because I don't want to have to deal with all the minor issues that will appear in the future.

mvucBmM0:
I've tried changing "download.list.layout" to "ui.torrent_list.layout" in
rpc::call_command("download.list.layout.set", (rpc::call_command_value("download.list.layout") + 1)%2);
in element_download_list.cc, but it didn't work.

rakshasa:
It's supposed to be "ui.torrent_list.layout.set".

mvucBmM0:
Yes.
rpc::call_command("ui.torrent_list.layout.set", (rpc::call_command_string("ui.torrent_list.layout") + 1)%2);
Doesn't work.

rakshasa:
Ahh... you need to look at the changes I made to the code... It's "full" and "compact", not 0 and 1.

rakshasa:
Use the "branch" command to make it work as you expect, or just add a C++ method that switches between modes.

mvucBmM0:
*note to self* I really need to learn C...

pyroscope:
BTW, why is this called "torrent_list" when really *everywhere* else it is called "download_list"?

This should be fixed before a release, since it's yet-another-dreaded-rename-of-commands waiting to happen.

PR #479, opened 2016-08-06
Expose the number of current HTTP connections

kannibalox:


rakshasa:
The word 'open' means you're trying to open a http connection, which is why the API calls it 'active'. Otherwise OK.

kannibalox:
No problem, did you want me to change all the names or just the new one?

rakshasa:
No 'open' has a specific, and commonly used, meaning in those cases as it's not just the word 'open' by itself.

I'd say "current_open" is the better option as it has some public usage and fits with the API in general. Changing the 'active' methods called by the API is up to you, however a patch for that would be nice.

kannibalox:
Updated with the name change, I don't feel the need to change the other two if you don't.

PR #493, opened 2016-08-25
Include rtorrent Wiki pages in doc/wiki directory

chros73:
Include rtorrent Wiki pages in `doc/wiki` directory.

**A. Theory:**

I. There's a neat way to include all the wiki pages in the source itself:
- use 'subtree' feature of git ( this is not 'subtree merge', [read more](https://hpc.uni.lu/blog/2014/understanding-git-subtree/))
  - easy to setup
  - easily maintainable
  - support branches

II. Since rtorrent wiki is community driven, that's how it should be used:
- every wiki pages should be created/edited online or using the main Wiki repo: https://github.com/rakshasa/rtorrent.wiki.git (not creating/editing files in the doc/wiki directory!)
- the `wiki` subtree has to be frequently updated, especially before creating a new release

**B. Usage:**

I. The `wiki` subtree inside rtorrent project has been created by the following commands:
```
git remote add -f wiki https://github.com/rakshasa/rtorrent.wiki.git
git subtree add --prefix doc/wiki wiki master --squash
git subtree pull --prefix doc/wiki wiki master --squash --message="Merge wiki updates into doc/wiki"
git push
```

II. Updating `wiki` subtree:
- select the branch you want to be updated (e.g. `master`)
```
git subtree pull --prefix doc/wiki wiki master --squash --message="Merge wiki updates into doc/wiki"
git push
```


rakshasa:
What exactly is this trying to do that the other wiki git repository doesn't handle?

Mainly the issue is that it seems to import all the wiki files into the main rtorrent repository, rather than being e.g. a git submodule.

chros73:
It will become a sort of type of submodule with this approach.
Goals:
- include all the docs (from wiki) in the project itself by moving all the existing (non-script) docs to wiki (you don't need to "maintain" a separate wiki repo)
- "keep track" of the wiki docs and project, so releases will contain the current state of wiki (e.g. commands can be changed in time, as it was before)

Both goals can be accomplished different ways (e.g. manually copying the whole wiki content into main repo; linking releases to certain wiki commits/tags), but I think this is the most useful way to do it.
If you don't like it, feel free to reject it.

rakshasa:
It seems to potentially cause issues with wiki-related commits being interspersed into what should be a development-oriented revision history.

Including wiki files and such should only really be done when e.g. making packages or installing, etc.

chros73:
> ... wiki-related commits being interspersed ...

That's correct. It a signal that docs were updated.

> Including wiki files and such should only really be done when e.g. making packages or installing, etc.

That's another approach, which is fine as well. :)

PR #562, opened 2017-02-09
Fixed including ncurses header based on autoconf

theirix:
This PR includes an ncurses header based on autoconf macro family `HAVE_NCURSES*`. It fixes an error happened when only `libncursesw5-dev` was installed on Debian 9.


rakshasa:
Fixed in dd2a4cc15192e712ee97b787967fa9fd390dbdea.

PR #575, opened 2017-03-06
Fix partially done downloads and choke groups

chros73:
Depends on:
- https://github.com/rakshasa/libtorrent/issues/147 (Fix partially done downloads and choke groups)

For more info:
- #564 (Bug in handling choke groups)
- https://github.com/rakshasa/rtorrent/issues/144#issuecomment-282481528 (rtorrent 0.9.3/0.13.3: Issues with "partially" downloaded torrent. Never finishes, completion percentage)

Also closes:
- #348 (stuck upload slots)

PR #604, opened 2017-05-15
Fix honoring system.file.allocate.set=1 rtorrent config setting

chros73:
Depends on: https://github.com/rakshasa/libtorrent/pull/152
Probably clashes with: https://github.com/rakshasa/rtorrent/pull/575 (it depends on it)

With the previous patch rtorrent could be easily crashed:
- e.g. when a magnet link was loaded, started and there wasn't enough disk space for it

**Modify fixing honoring `system.file.allocate.set=1` patch:**
- do not allow to start a torrent if there's not enough disk space for it
  - notify the user about this
- modify priorities of files as well in `libtorrent`
- add 2 new file property in `libtorrent`: 
  - `is_fallocatable` : if a file has `flag_fallocate` flag
  - `is_fallocatable_file` : if a file has `flag_fallocate` and `flag_resize_queued` flags
- modify `DownloadList::create()` to take into account `system.file.allocate` setting
- modify `d.update_priorities` to be able to pass  along file flags
- refactor `DownloadList::resume()` a bit

**New commands in rtorrent:**
- `d.is_enough_diskspace` : boolean
- `d.allocatable_size_bytes` : size in bytes needed to create the download
- `f.is_fallocatable` ,  : boolean, returns true if a file has `flag_fallocate` flag
- `f.is_fallocatable_file` : boolean, returns true if a file both has `flag_fallocate` and `flag_resize_queued` flags
- `f.set_fallocate`, `f.unset_fallocate` : setters for `flag_fallocate`

**Additional fix:**
- fix `free_diskspace()` method in `libtorrent` to report back proper size even if a download is stopped

**Note:**
Unfortunately, this patch isn't complete yet, rtorrent can be still crashed:
- load a multi file torrent that is larger than free disk space but not start it
- deselect some files to be able to start download and start it
  - file allocation will be immediate
- now select some previously deselected files that won't fit on the disk
  - rtorrent will crash

### DIFF
--- a/src/command_download.cc
+++ b/src/command_download.cc
@@ -64,6 +64,7 @@
 #include "core/download_store.h"
 #include "core/manager.h"
 #include "rpc/parse.h"
+#include "rpc/parse_commands.h"
 
 #include "globals.h"
 #include "control.h"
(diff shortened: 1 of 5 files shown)


chros73:
@rakshasa , I'll try to do something about the 2nd (incomplete) part this week. If I'll won't be able to figure out a good solution then I'll leave this with you (it's getting more and more complicated :) ).
Thanks

trim21:
I think I have fixed this, this can be closed now

PR #480, opened 2016-08-07
Display values of multiple throttle.up and throttle.down in the first part of status bar

chros73:
Display values of multiple throttle.up and throttle.down in the first part of status bar, also change name of config directive.

Currently the beginning of status bar looks like this:
`[Throttle 500/1500 KB]  [Rate: 441.6/981.3 KB]`

Include the max limit of the throttles, the main upload rate and the upload rate of the throttles (in this order); and main download rate and the download rate of the selected throttles.

Possible cases look like these:

```
[Throttle 200 / 500 KB] [Rate 107.4 / 298.6 KB]
[Throttle 200(114) / 500 KB] [Rate 107.0(1.0|105.9) / 307.6 KB]
[Throttle 200 / 500(250) KB] [Rate 124.7 / 298.2(298.2|0.0) KB]
[Throttle 200(114) / 500(250) KB] [Rate 115.9(1.7|114.2) / 333.9(333.9|0.0) KB]

[Throttle 3500(75|25) / 8000 KB] [Rate 1745.5(1635.1|85.0|25.4) /   7.3 KB]
```
- make the displayable throttle names configurable via config file: `ui.status.throttle.up.set` and `ui.status.throttle.down.set`
- original code is refactored
- throttle.up and throttle.down values can be specified simultanously (as in the last example)
- name of throttles must be specified without surrounding quotes (the setting no longer a `string` but a `list`)
  - `ui.status.throttle.up.set = slowup , tardyup`
  - `ui.status.throttle.down.set = slowup`
- limitation is that every group (there are 4 possible groups) can contain the following number of characters (it leaves space for at least 5 throttles to be displayed):
  - limit: 40 chars
  - rate: 50 chars
- don't display this extra info of a given throttle in the following cases:
  - there isn't any throttle name as the config variables suggest or the given name is "NULL"
  - the throttle is not throttled (=0)
  - the global upload/download is not throttled (=0) (the throttle won't be taken into account in this case)

Refers to: https://github.com/rakshasa/rtorrent/pull/447 , https://github.com/rakshasa/rtorrent/issues/564#issuecomment-285867371

PR #615, opened 2017-06-06
Implement input (command) history with categories

chros73:
Implement input (command) history for input prompts (e.g. with `^X`), as in bash (without searching capability) with categories (mainly because of the upcoming temp-filter feature #519):
- all the different input types have their own history:
  - `load_default`, `load_modified`, `change_directory`, `command`, `filter`

**Functionality:**
- entry won't be added to a category history if it's:
  - empty
  - the same as the previous entry
- earlier entry will be updated if it's changed while moving "away" from it and not "applying" it
- last (temp) entry is remembered (until the next prompt)
- history is saved during `session.save` into session directory (if it exists) as `rtorrent.input_history`
- history is loaded at startup (if it's available)
- latest entries are kept during resizing history categories if it's smaller than original

**Added keyboard shortcuts at input prompt:**
- `up arrow`, `^p` : prev in history category
- `down arrow`, `^n` : next in history category
- `ESC`, `^[`, `^g` (latter was already defined): exit from input prompt
  - `ESC` was "used" to mimic `ALT` at input prompt back in the days but it isn't used there anymore, hence the removal of the code from `TextInput`.

**Added commands:**
- `ui.input.history.size.set` : set size of every history category to this value
  - default is: `99`
- `ui.input.history.size` : get size of a history category
- `ui.input.history.clear` : clear all the history categories

**Sample `rtorrent.input_history` file:**
```
1|301
1|302
1|303
3|/mnt/rtorrent/various/
3|/mnt/rtorrent/unsafe/
4|session.save=
4|uptime=
4|print=ok
4|method.insert = try_to_print, simple|const, "print=9"
5|book.*pdf
```

**Existing keyboard shortcuts at input prompt (for reference):**
- `^L` : force redraw display
- `^Q` : normal shutdown
- `TAB` : path completion at `change_directory` prompt
- `BACKSPACE`, `^H`
- `DEL`
- `left arrow`, `^B` : move 1 char to left
- `right arrow`, `^F` : move 1 char to right
- `HOME`, `END`
- `^U` : erase from cursor to beginning
- `^K` : erase from cursor to end

**Implementation notes:**
- use prefilled ring buffers for each category in history
- keep track of insert pointers for each category in history
- it's enough to store one last (temp) entry and get pointer for all the categories in history

**Notes:**
@rakshasa,.
- probably you'll get merge issues with other pull requests (e.g. #519 , #480, etc.)
- [INPUT_FILTER](https://github.com/rakshasa/rtorrent/pull/519/files#diff-44fab6581e2392a72b26776ddf50485fR90) from #519 needs to put between `INPUT_COMMAND` and `INPUT_EOI`

### DIFF
--- a/src/command_ui.cc
+++ b/src/command_ui.cc
@@ -555,8 +555,9 @@ initialize_command_ui() {
   CMD2_ANY       ("ui.current_view",       std::bind(&cmd_ui_current_view));
   CMD2_ANY_STRING("ui.current_view.set",   std::bind(&cmd_ui_set_view, std::placeholders::_2));
 
+  CMD2_ANY         ("ui.input.history.size",     std::bind(&ui::Root::set_input_history_size, control->ui()));
   CMD2_ANY_VALUE_V ("ui.input.history.size.set", std::bind(&ui::Root::set_input_history_size, control->ui(), std::placeholders::_2));
-  CMD2_ANY_V       ("ui.input.history.clear", std::bind(&ui::Root::clear_input_history, control->ui()));
+  CMD2_ANY_V       ("ui.input.history.clear",    std::bind(&ui::Root::clear_input_history, control->ui()));
 
   // TODO: Add 'option_string' for rtorrent-specific options.
   CMD2_VAR_STRING("ui.torrent_list.layout", "full");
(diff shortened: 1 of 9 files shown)


chros73:
- change default `ui.input.history.size.set` to `99` (`30` was small in practice)
- add `ui.input.history.size` as a getter command for the sake of completeness

Note: original post is also changed.

chros73:
Note, 1 building job failed due to apt error (nothing has to do with the modded code), and since I can't re-trigger it ... :)

rakshasa:
Merged.

Issue #374, opened 2016-01-19
Clarify how to use or fix directory.watch.added


akintoye:
Your wiki page [TORRENT Watch directories](https://github.com/rakshasa/rtorrent/wiki/TORRENT-Watch-directories) states that one can use inotify on Linux for watching directories:

> `directory.watch.added = "~/Download/watch/", load.start`
On Linux you may use inotify to watch a directory, and the command is called with the the full path of new files as the first argument

However, I can't figure out how to use this command. 

Having 
```
directory.watch.added = "~/Download/watch/", load.start
```
in the rtorrent.rc, starting rtorrent results in the error:
```
rtorrent: Failed to parse command line option: Error in option file: ./.rtorrent.rc:43: Command "directory.watch.added" does not exist.
```
I figured, one should use it like the other commands `start_tied`, `stop_untied`, etc., so I put 
```
schedule = inotify_watch_directory,0,0,"directory.watch.added=~/Download/watch/,load.start"
```
into the rc. The, rtorrent starts up, but prints:
```
(11:01:37) Scheduled command failed: inotify_watch_directory: Command "directory.watch.added" does not exist.
```

I'm on version 0.9.6-3 of the arch linux package (Rakshasa's BitTorrent client version 0.9.6.)

chros73:
1. about directory.watch.added :
It wasn't available in 0.9.4 when I tried it half a year ago. Another issue can be with it, if you want to pass on additional parameters as you can do with the traditional watch directories (I'm using them). And since you can schedule watchers 10 seconds it's not really a problem.

2. I tried to use inotify for my queue script (outside of rtorrent) but I also gave it up:
I couldn't figure it out how you can batch process several torrent files at a given time (e.g. when the first come in then wait for e.g. 40 seconds for possible others). I've found a python solution for it, but it wasn't so relaible. So I'm just using cron for this one (5 mins).

pyroscope:
https://github.com/rakshasa/rtorrent/wiki/TORRENT-Watch-directories/_compare/23a3383828fbddeb7f01fa81984db00f2f9cf1f1...887d6d23bcce921a8c1cc08135f0125a4a29218e

Can be closed.

chros73:
>  Another issue can be with it, if you want to pass on additional parameters as you can do with the traditional watch directories

How can we do this? Wiki says its possible but I think it isn't with the current implementation.

pyroscope:
Of course it doesn't if you quote the commands. That is what quoting does.

[...]

PR #616, opened 2017-06-10
Modify directory.watch.added and add directory.watch.removed command

chros73:
One of the advantages of using "inotify" to load and remove downloads by torrent meta files is the reduced  file system IO (apart from it's instant).

Refers to: https://github.com/rakshasa/rtorrent/issues/374

**A. Modify `directory.watch.added`**
Allow it to accept multiple commands (as `load.*` commands).
There are multiple ways to achieve this but since there's no "shortcut" (the directory (that is watched) for inotify has to be separated), the final command (that is called on an "add" event) is reconstructed as a line of config.
- 1st param: the directory that will be watched
- 2nd param: the name of the main command that will be called if an "add" event is triggered (`load.*` commands)
- rest of the params: extra commands that will be passed as arguments to the above specified main command
  - if an extra command include commas (`,` parameter separator) then it needs to be included inside quotes (`"`)
```ini
directory.watch.added = (cat,(cfg.dir.meta_downl),unsafe/),   load.start,  "d.attribs.set=unsafe,,1", print=loadedunsafe
```

**B. Add `directory.watch.removed`**
Similarly to loading event, add ability to handle removing events of meta files as `*_untied` commands:
- `*_untied` commands can't deal with a download individually: they go through and check all the downloads (they can generate lot of IO, especially if they are used with `schedule` command)
- `directory.watch.removed` handles this individually

Since it doesn't need any additional commands, it accepts multiple directories as parameters.
- 1st param: the command to call when a meta file is removed, only 3 commands are supported
  - `d.stop`, `d.close`, `d.erase` 
- rest of the params: comma separated list of the directories that will be watched
```ini
directory.watch.removed = d.erase, (cat,(cfg.dir.meta_compl),various/), (cat,(cfg.dir.meta_compl),unsafe/)
```

**C. Rename `*tied` commands**
Rename untied commands and add redirects for the new names to preserve compatibility:
- `start_tied` -> `tied.start`
- `stop_untied` -> `untied.stop`
- `close_untied` -> `untied.close`
- `remove_untied` -> `untied.remove`

**D. Limitations**
- a given directory can only be specified once with either `directory.watch.added` or `directory.watch.removed`
- torrent meta files that are "added" / "removed" into/from watched directories will be discarded if rtorrent wasn't run at the time of adding / removing
  - it's a good idea to run the normal scheduled "watch" command once after rtorrent is started
- probably it won't work well with network file systems (NFS, Samba, etc.)

**E. `directory.watch.*` commands in action**
Let's assume the following:
- we have multiple watch directories with which we want to use different properties for downloads with a parameterised command (like this [d.attribs.set](https://github.com/chros73/rtorrent-ps_setup/blob/develop/ubuntu-14.04/home/chros73/.rtorrent.rc#L53))
- we want to provide the ability of switching inotify support for users in config
- we also want to watch directories for removed metafiles
  - the meta files is moved by rtorrent (internally) when a download is finished

```ini
# Defining directory constants
method.insert = cfg.dir.main,       string|const|private, (cat,"/mnt/Torrents/")
method.insert = cfg.dir.sub,        string|const|private, (cat,(cfg.dir.main),".rtorrent/")
method.insert = cfg.dir.meta_downl, string|const|private, (cat,(cfg.dir.sub),".downloading/")
method.insert = cfg.dir.meta_compl, string|const|private, (cat,(cfg.dir.sub),".completed/")

# Whether to use inotify for loading/removing torrent (meta) files from watch directories
method.insert = cfg.inotify.use, value|const|private, 1

# Whatch dir definitions for loading metafiles using inotify
branch = ((cfg.inotify.use)), ((directory.watch.added, (cat,(cfg.dir.meta_downl),rotating/), load.start,  "d.attribs.set=rotating,1,1"))
branch = ((cfg.inotify.use)), ((directory.watch.added, (cat,(cfg.dir.meta_downl),fullseed/), load.start,  "d.attribs.set=fullseed,1,"))
branch = ((cfg.inotify.use)), ((directory.watch.added, (cat,(cfg.dir.meta_downl),unsafe/),   load.start,  "d.attribs.set=unsafe,,1"))
branch = ((cfg.inotify.use)), ((directory.watch.added, (cat,(cfg.dir.meta_downl),various/),  load.start,  "d.attribs.set=various,,"))
branch = ((cfg.inotify.use)), ((directory.watch.added, (cat,(cfg.dir.meta_downl),load/),     load.normal, "d.attribs.set=unsafe,,1"))

# Fire up only once (after rtorrent is started) the normal scheluded load watch dirs as well even if inotify is used: load added metafiles when rtorrent hasn't been run
branch = ((cfg.inotify.use)), ((schedule2, watch_dir_1,  85, 0, ((load.start, (cat,(cfg.dir.meta_downl),rotating/*.torrent), "d.attribs.set=rotating,1,1"))))
branch = ((cfg.inotify.use)), ((schedule2, watch_dir_2,  86, 0, ((load.start, (cat,(cfg.dir.meta_downl),fullseed/*.torrent), "d.attribs.set=fullseed,1,"))))
branch = ((cfg.inotify.use)), ((schedule2, watch_dir_3,  87, 0, ((load.start, (cat,(cfg.dir.meta_downl),unsafe/*.torrent),   "d.attribs.set=unsafe,,1"))))
branch = ((cfg.inotify.use)), ((schedule2, watch_dir_4,  88, 0, ((load.start, (cat,(cfg.dir.meta_downl),various/*.torrent),  "d.attribs.set=various,,"))))
branch = ((cfg.inotify.use)), ((schedule2, watch_dir_5,  89, 0, ((load.normal,(cat,(cfg.dir.meta_downl),load/*.torrent),     "d.attribs.set=unsafe,,1"))))

# Whatch dir definitions for removing metafiles using inotify
branch = ((cfg.inotify.use)), ((directory.watch.removed, d.erase, (cat,(cfg.dir.meta_compl),various/), (cat,(cfg.dir.meta_compl),unsafe/), (cat,(cfg.dir.meta_compl),rotating/), (cat,(cfg.dir.meta_compl),fullseed/), (cat,(cfg.dir.meta_compl),load/)))

# Whatch dir definitions for loading metafiles NOT using inotify
branch = ((not, (cfg.inotify.use))), ((schedule2, watch_dir_1,  5, 10, ((load.start, (cat,(cfg.dir.meta_downl),rotating/*.torrent), "d.attribs.set=rotating,1,1"))))
branch = ((not, (cfg.inotify.use))), ((schedule2, watch_dir_2,  6, 10, ((load.start, (cat,(cfg.dir.meta_downl),fullseed/*.torrent), "d.attribs.set=fullseed,1,"))))
branch = ((not, (cfg.inotify.use))), ((schedule2, watch_dir_3,  7, 10, ((load.start, (cat,(cfg.dir.meta_downl),unsafe/*.torrent),   "d.attribs.set=unsafe,,1"))))
branch = ((not, (cfg.inotify.use))), ((schedule2, watch_dir_4,  8, 10, ((load.start, (cat,(cfg.dir.meta_downl),various/*.torrent),  "d.attribs.set=various,,"))))
branch = ((not, (cfg.inotify.use))), ((schedule2, watch_dir_5,  9, 10, ((load.normal,(cat,(cfg.dir.meta_downl),load/*.torrent),     "d.attribs.set=unsafe,,1"))))

# Removes torrents from client when its metafile (torrent file) has been deleted manually or by a script. If inotify isn't used: Run it every 5 seconds, otherwise run only once after rtorrent is started.
branch = ((cfg.inotify.use)), ((schedule2, untied_torrents, 99, 0, ((untied.remove)))), ((schedule2, untied_torrents, 8, 10, ((untied.remove))))
```

### DIFF
--- a/src/command_events.cc
+++ b/src/command_events.cc
@@ -308,24 +308,88 @@ d_multicall(const torrent::Object::list_type& args) {
 }
 
 static void
-call_watch_command(const std::string& command, const std::string& path) {
-  rpc::commands.call_catch(command.c_str(), rpc::make_target(), path);
+call_watch_command_added(const core::Manager::command_list_type& args, const std::string& path) {
+  if (args.size() < 1)
+    throw torrent::input_error("Too few arguments.");
+
+  std::string rpc_command_str;
+  core::Manager::command_list_type::const_iterator argsItr = args.begin();
+
+  std::string command = *argsItr;
+
+  rpc_command_str = command + (*command.rbegin() != '=' ? "=" : "") + path;
+
+  while (++argsItr != args.end())
+    rpc_command_str += ", \"" + *argsItr + "\"";
+
+  rpc::parse_command_single_std(rpc_command_str);
 }
 
 torrent::Object
 directory_watch_added(const torrent::Object::list_type& args) {
-  if (args.size() != 2)
+  if (args.size() < 2)
     throw torrent::input_error("Too few arguments.");
-
-  const std::string& path = args.front().as_string();
-  const std::string& command = args.back().as_string();
 
   if (!control->directory_events()->open())
     throw torrent::input_error("Could not open inotify:" + std::string(rak::error_number::current().c_str()));
 
-  control->directory_events()->notify_on(path.c_str(),
+  core::Manager::command_list_type commands;
+  torrent::Object::list_const_iterator argsItr = args.begin();
+
+  const std::string& path = argsItr->as_string();
+
+  while (++argsItr != args.end())
+    commands.push_back(argsItr->as_string());
+
+  control->directory_events()->notify_on(path,
                                          torrent::directory_events::flag_on_added | torrent::directory_events::flag_on_updated,
-                                         std::bind(&call_watch_command, command, std::placeholders::_1));
+                                         std::bind(&call_watch_command_added, commands, std::placeholders::_1));
+  return torrent::Object();
+}
+
+static void
+call_watch_command_removed(const std::string& command, const std::string& path) {
+  if (command != "d.stop" && command != "d.close" && command != "d.erase")
+    return;
+
+  for (core::DownloadList::iterator itr = control->core()->download_list()->begin(); itr != control->core()->download_list()->end(); ++itr) {
+    const std::string& tiedToFile = rpc::call_command_string("d.tied_to_file", rpc::make_target(*itr));
+
+    if (!tiedToFile.empty() && rak::path_expand(path) == rak::path_expand(tiedToFile)) {
+
+      if (command == "d.stop" && rpc::call_command_value("d.state", rpc::make_target(*itr)) != 0) {
+        rpc::parse_command_single(rpc::make_target(*itr), "d.try_stop=");
+      } else if (command == "d.close" && rpc::call_command_value("d.ignore_commands", rpc::make_target(*itr)) == 0) {
+        rpc::parse_command_single(rpc::make_target(*itr), "d.try_close=");
+      } else if (command == "d.erase") {
+        // Need to clear tied_to_file so it doesn't try to delete it.
+        rpc::call_command("d.tied_to_file.set", std::string(), rpc::make_target(*itr));
+        control->core()->download_list()->erase(itr);
+      }
+
+      break;
+    }
+  }
+}
+
+torrent::Object
+directory_watch_removed(const torrent::Object::list_type& args) {
+  if (args.size() < 2)
+    throw torrent::input_error("Too few arguments.");
+
+  torrent::Object::list_const_iterator argsItr = args.begin();
+  std::string command = argsItr->as_string();
+
+  if (command != "d.stop" && command != "d.close" && command != "d.erase")
+    throw torrent::input_error("directory.watch.removed command only supports d.stop , d.close , d.erase commands.");
+
+  if (!control->directory_events()->open())
+    throw torrent::input_error("Could not open inotify:" + std::string(rak::error_number::current().c_str()));
+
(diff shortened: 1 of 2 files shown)


pyroscope:
The limitation of "removed" is arbitrary, and it is not very flexible. 

The better / more consistent way would be to fire an `event.download.untied`, and then that multi-method can take care of things – it would be called on the item(s) tied to the removed file.

And then the next logical step is call it `directory.watch.tied_to_files`, and also trigger `event.download.tied`.

pyroscope:
And while looking at that `d.multicall.check.tied_to_files` name, `d.multicall.check.base_path` would be nice too – fire `event.download.orphaned` when an item's data is gone. Obvious handlers would be `d.erase`, or `d.close` and maybe also `d.views.push_back_unique=orphaned`. Yes, events are *that* flexible.

pyroscope:
See https://github.com/rakshasa/rtorrent/issues/374#issuecomment-398242223 about why you do not need more than the handler name as an argument.

chros73:
A. `directory.watch.removed` : the idea was to provide similar method as `directory.watch.added`.

> The limitation of "removed" is arbitrary, and it is not very flexible

I agree, although it's working fine:
- a given directory can only be specified once with either `directory.watch.added` or `directory.watch.removed`

About the events: it's an interesting idea, but somebody else has to implement it :) .
Right now similar things can be achieved with the allowed commands and their corresponding events (erased, stopped, etc.) and a bit of scripting (if it's necessary).

rakshasa:
New watch command has been added, if this PR is still relevant it needs updating.

PR #648, opened 2017-10-03
Simplify curl dual stack IPv4 / IPv6 handling

anthonyryan1:
curl implements the RFC 6555 "Happy Eyeballs" algorithm, meaning a
manual fallback to IPv6 is unnecessary and arguably detrimental.

If we leave curl to it's own devices with the request, it will attempt
to connect on both IPv4 and IPv6, then use whichever connection is
going to be more reliable.

### DIFF
--- a/src/core/download_factory.h
+++ b/src/core/download_factory.h
@@ -77,7 +77,7 @@ public:
 
   bool                get_start() const     { return m_start; }
   void                set_start(bool v)     { m_start = v; }
-  
+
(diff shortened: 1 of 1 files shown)


rakshasa:
I'm working on the feature-bind branch in order to add better control of when and when to use ipv4 vs. ipv6.

If you, and others, want to help implement ipv6 support I hope you'd look at <https://github.com/rakshasa/rtorrent-vagrant>.

Issue #515, opened 2016-10-25
Race condition between CommandScheduler insert & call_item


anthonyryan1:
When a new CommandScheduleritem is inserted into the CommandScheduler with the same key as an already existing item, the previous item is deleted (freeing memory) before replacing it with a new item.

Unfortunately, there's a race condition here when insert runs at the same time work is being performed (for instance call_item) on another thread.

I've included a stack running on 0.9.6 demonstrating this problem. I only see two obvious solutions:

1. Allowing duplicate keys in the CommandScheduler (eliminating the delete on insert).
2. Using proper locking / synchronization to coordinate manipulation of the CommandScheduler vector Items without racing.

I'd love some thoughts about which solution would fit in better here.

--------

READ of size 8 thread 0 (rtorrent main)

0. rak::timer::operator!=(rak::timer const&) const ../../rak/timer.h:83
1. rak::priority_item::is_queued() const ../../rak/priority_queue_default.h:63
2. rpc::CommandSchedulerItem::is_queued() const rtorrent/src/rpc/command_scheduler_item.h:54
3. rpc::CommandSchedulerItem::enable(rak::timer) rtorrent/src/rpc/command_scheduler_item.cc:54
4. rpc::CommandScheduler::call_item(rpc::CommandSchedulerItem*) rtorrent/src/rpc/command_scheduler.cc:117
5. std::tr1::_Mem_fn<void (rpc::CommandScheduler::*)(rpc::CommandSchedulerItem*)>::operator()(rpc::CommandScheduler*, rpc::CommandSchedulerItem*) const /usr/lib/gcc/x86_64-pc-linux-gnu/4.9.3/include/g++-v4/tr1/functional:585
[...]

PR #681, opened 2018-01-24
Fix race condition in CommandScheduler

chros73:
Patched by @pyroscope.

Closes: #515

### DIFF
--- a/src/rpc/command_scheduler.cc
+++ b/src/rpc/command_scheduler.cc
@@ -63,15 +63,18 @@ CommandScheduler::insert(const std::string& key) {
   if (key.empty())
     throw torrent::input_error("Scheduler received an empty key.");
 
+  CommandSchedulerItem* current = new CommandSchedulerItem(key);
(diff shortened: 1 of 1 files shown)


rakshasa:
This isn't the fix you are looking for.

It doesn't address the race condition as the scheduler isn't a thread-safe container, and at best you hide the issue a bit.

The scheduler is only ever supposed to be called while in the main thread, or updated while the global lock is being held by a particular thread.

If there's an issue it might be due to the worker thread starting in main.cc before the main thread.

chros73:
> It doesn't address the race condition ... If there's an issue it might be due to the worker thread starting in main.cc before the main thread.

Ok, thanks for the explanation.
I modified "Closes" to "Relates to" in the description.

> ... at best you hide the issue a bit.

Well, it prevents crashing. You can modify the title of this pull request if you still feel that this is useful.

rakshasa:
To debug this we probably should add a check to see if the thread calling the command holds the global lock.

rakshasa:
Worker thread no longer executes commands, instead they are passed to main thread.

PR #722, opened 2018-05-25
Update rtorrent.rc

g0tmi1k:
Useful: 
- https://media.readthedocs.org/pdf/rtorrent-docs/latest/rtorrent-docs.pdf
- https://rtorrent-docs.readthedocs.io/en/latest/ 
- https://github.com/rakshasa/rtorrent/wiki/rTorrent-0.9-Comprehensive-Command-list-(WIP)
- https://github.com/rakshasa/rtorrent/blob/master/src/command_local.cc

### DIFF
--- a/doc/rtorrent.rc
+++ b/doc/rtorrent.rc
@@ -1,9 +1,17 @@
 #############################################################################
-# This is an example resource file for rTorrent.
+# This is an (old) example resource file for rTorrent.
 # Copy to ~/.rtorrent.rc and enable/modify the options as needed.
 # Remember to uncomment the options you wish to enable.
 #
-# File: https://github.com/rakshasa/rtorrent/blob/master/doc/rtorrent.rc
+# See 'rtorrent.rc-example' for a newer, basic configuration.
+#
+#   Sample: https://github.com/rakshasa/rtorrent/wiki/CONFIG-Template
+# Complete: https://rtorrent-docs.readthedocs.io/en/latest/cmd-ref.html
+#   Useful: https://rtorrent-docs.readthedocs.io/en/latest/use-cases.html
+#   Manual: https://rtorrent-docs.readthedocs.io/en/latest/
+#  Convert: https://github.com/rakshasa/rtorrent/wiki/rTorrent-0.9-Comprehensive-Command-list-(WIP)
+# Handbook: https://media.readthedocs.org/pdf/rtorrent-docs/latest/rtorrent-docs.pdf
+#     File: https://github.com/rakshasa/rtorrent/blob/master/doc/rtorrent.rc
 #############################################################################
 
 # Maximum and minimum number of peers to connect to per torrent.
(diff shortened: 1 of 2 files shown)


pyroscope:
If you put the socket file into /tmp, i.e. a shared dir, then at least qualify it with the PID or the username. And a way better place is ``session.path``.

g0tmi1k:
@pyroscope Thanks for the suggestion. Updated.

pyroscope:
Also, if you don't use the modern wiki template, at least surgically remove the cancer that the horrible `stop_untied` schedule is. Problems caused by this are a FAQ.

g0tmi1k:
Removed `stop_untied` as your request @pyroscope .
I had a quick look at the FAQ (https://github.com/rakshasa/rtorrent/blob/master/doc/faq.xml), and ctrl+F didn't have anything about it, nor could I see anything in the wiki (https://github.com/rakshasa/rtorrent/wiki) about it.

I've added the "sample" config in (`rtorrent.rc-sample`) from the wiki, but would it be better to replace this `rtorrent.rc` completely with it?
I personally wouldn't, but it sounds like it may be helpful?
I wouldn't as it would be "nice" to see all the possible options there are to put in a config file (are any missing still?)

Also`feature-bind` branch is ahead of master but a bit (Jan 2018), which hasn't been updated in some time (Oct 2016). The pull request best going into master? [cc: @rakshasa]

pyroscope:
Look at reddit, there it is a FAQ, over several years.

g0tmi1k:
https://www.reddit.com/r/rtorrent/wiki/index

```
forbidden (reddit.com)
you are not allowed to do that

— wiki_disabled.
```

pyroscope:
For “all the possible options there are” see the common tasks page, https://rtorrent-docs.readthedocs.io/en/latest/use-cases.html, and https://rtorrent-docs.readthedocs.io/en/latest/cmd-ref.html. 

A starter config has to stick to the essentials, else you overwhelm people.

g0tmi1k:
@pyroscope I agree a starter config should be basic as it does overwhelm/confuse people.

I also believe there should be a place to go with all the options/configs listed in a single place (which is what I though `rtorrent.rc` was). Guessing the website is more up-to-date. Is it worth linking in either/both `rtorrent.rc*` files to that place so its easier for others to find?

chros73:
> The pull request best going into master?

That's correct, master branch.

pyroscope:
> Is it worth linking in either/both rtorrent.rc* files to that place so its easier for others to find?
The new one is, the old stinky one won't be.

g0tmi1k:
@pyroscope Added a ton of useful resources to the old one, pointing to the new style.
What else would be needed/recommended for the merge?

pyroscope:
Put anything in there not on common tasks on common tasks – that is much more discoverable and useful.

g0tmi1k:
@pyroscope Like how it is?

g0tmi1k:
Fix merge conflicts (due to #677)

g0tmi1k:
🎉

PR #744, opened 2018-06-15
add 'compare' command / operator

pyroscope:
`compare = <order>, <sort_key1>=[, ...]`

Compares two items like `less=` or `greater=`, but allows to compare by several different sort criteria, and ascending or descending order per given field.

The first parameter is a string of order indicators, either `aA+` for ascending or `dD-` for descending. The default, i.e. when there's more fields than indicators, is ascending. Field types other than value or string are treated as equal (or in other words, they're ignored).

If all fields are equal, then items are ordered in a random, but stable fashion.

Configuration example:

    # VIEW: Show active and incomplete torrents (in view #9) and update every 20 seconds
    #       Items are grouped into complete, incomplete, and queued, in that order.
    #       Within each group, they're sorted by upload and then download speed.
    view.sort_current = active,"compare=----,d.is_open=,d.complete=,d.up.rate=,d.down.rate="
    schedule = filter_active,12,20,"view.filter = active,\"or={d.up.rate=,d.down.rate=,not=$d.complete=}\" ;view.sort=active"

### DIFF
--- a/src/command_ui.cc
+++ b/src/command_ui.cc
@@ -284,6 +284,56 @@ torrent::Object apply_equal(rpc::target_type target, const torrent::Object::list
 }
 
 torrent::Object
+apply_compare(rpc::target_type target, const torrent::Object::list_type& args) {
+  if (!rpc::is_target_pair(target))
+    throw torrent::input_error("Can only compare a target pair.");
+
+  if (args.size() < 2)
+    throw torrent::input_error("Need at least order and one field.");
+
+  torrent::Object::list_const_iterator itr = args.begin();
+  std::string order = (itr++)->as_string();
+  const char* current = order.c_str();
+
+  torrent::Object result1;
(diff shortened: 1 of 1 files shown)


src/command_ui.cc
```diff
@@ -283,6 +283,56 @@ torrent::Object apply_equal(rpc::target_type target, const torrent::Object::list
   return result.is_value() ? result.as_value() == 0 : (int64_t)false;
 }
 
+torrent::Object
+apply_compare(rpc::target_type target, const torrent::Object::list_type& args) {
+  if (!rpc::is_target_pair(target))
+    throw torrent::input_error("Can only compare a target pair.");
+
+  if (args.size() < 2)
```
rakshasa: There's no need to require this, make it work with zero compare fields.

Also, the order field should allow empty object or no args at all.

PR #763, opened 2018-07-27
travis: Fix test in wrong path, and jobs parallelization

MindTooth:
Cut a lot out of my previous PR.  Now it just fixes the wrong path, and moves the parallelization into an environment variable.

Closes #728.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,9 @@
 language: cpp
 
+env:
+  global:
+    - MAKEFLAGS="-j 12"
+
 matrix:
   include:
   - compiler: clang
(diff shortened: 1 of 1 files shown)


rakshasa:
It now seems to fail travis builds.

MindTooth:
Yes, because it now tests the `rtorrent` source code, not building `libtorrent` twice.  Unless that was the goal from the start.

I thought it were pulled & build `libtorrent` only as a dependency, and the `make check` were for `rtorrent`.  Current form will never test the `rtorrent` source AFAIK.

MindTooth:
Ended up putting in on multiple lines.  Also, maybe an idea to include newer toolchains?  Know that it won't compile on macOS, as it's been removed from Homebrew.

MindTooth:
Ended up putting in on multiple lines. Also, maybe an idea to include newer toolchains? Know that it won't compile on macOS, as it's been removed from Homebrew.

MindTooth:
Figured I would give this another try.  After some testing inside a clean [Ubuntu Trusty VM](https://app.vagrantup.com/ubuntu/boxes/trusty64), it compiles just fine after I've installed the dependancies.  But when I use a clean file on Travis-CI, it breaks on `make check`.  I suspect Travis-CI is at falt.  I'll be doing some more research what might cause this.

The `.travis.yml` file:
```yml
language: cpp

addons:
  apt:
    packages:
      - libcppunit-dev
    update: true

# TODO: Use the same branch name if libtorrent has it.
before_install:
 - git clone https://github.com/rakshasa/libtorrent.git
 - |
   pushd libtorrent \
   && ./autogen.sh \
   && ./configure \
   && make \
   && sudo make install \
   && popd

script:
- ./autogen.sh && ./configure && make
- if [ ! $SKIP_CHECK ]; then make check; fi
```

MindTooth:
Woot!  I find the issue; needed to add `--prefix=/usr` to libtorrent configure.

Now it succeeds with all compilers.

MindTooth:
Been thinking about also adding more environments to the matrix.  The current toolchains is a bit out of date.  The default Ubuntu image is now defaulting to xenial (16.04 LTS), and I would believe that the default toolchain have changed.

_(If the idea is to build libtorrent twice, I understand, and I'll close this PR.)_

Issue #239, opened 2014-09-25
cannot find readable session directory. check permissions


Apathialol:
Lubuntu 14.04
I'm trying to run rTorrent as a service on startup but it's refusing to work.

I installed rTorrent using apt-get, am using the init script found here: ]http://libtorrent.rakshasa.no/downloads/rtorrent_init_script.sh and modified it appropriately (changed username and added init info to the top)
Everything has proper permissions.

Running the commands "screen rtorrent" and "sudo screen rtorrent" work just fine.
But if I try running "service rtorrent start" I'm met with the error  "cannot find readable session directory /home/servu/.rtorre. check permissions"
I had to create a symlink for /home/servu/.rtorrent.rc, which is what the config should be, so I'm not even sure where it's getting .rtorre from.

So is there something I'm doing wrong or what?

waveletlet:
I was having this problem with the script as well. I found that it was due to the change in config file option naming after 0.9. The session directory is now defined by "session.path.set" instead of "session", and the old init script won't find the "session.path.set" option.

PR #794, opened 2018-10-18
 Edit getsession() for 0.9 config format 

waveletlet:
Changed the regex in the getsession() function. I only tested the .sh version as I am on a system without bash.

Resolves https://github.com/rakshasa/rtorrent/issues/239


kannibalox:
`session` is still a valid config field. While I'll encourage people to use the new syntax, I don't think we should break things for the non-deprecated commands.

PR #897, opened 2019-07-21
Fix a few warnings.

neheb:


rakshasa:
I'm not going to break all the patches out there just to rename NULL.

Later I'll use clang-tidy to update all the code.

PR #898, opened 2019-07-25
Added log.append_file command

Nicholi:
If https://github.com/rakshasa/libtorrent/pull/195 gets approved, this makes use of new log_open_file_output_append function.

Meant to have identical behavior to "log.open_file", but append file when opening instead of truncate/overwrite.


Nicholi:
Updated with feedback from https://github.com/rakshasa/libtorrent/pull/195.
Also included log.append_gz_file command now.

PR #937, opened 2019-11-25
few missing sub-commands of network and throttle for the RPC interface

Trogious:
Adding few missing commands to the RPC interface:
```
network.total_handshakes
network.open_files
throttle.max_unchoked_uploads
throttle.max_unchoked_downloads
```


rakshasa:
Looks good, could you fix the spacing so the argument align with the rest of the code?

Trogious:
Done.

PR #967, opened 2020-02-22
new method close_low_diskspace.normal

pyroscope:
Skip downloads with prio=3 (high) when checking for disk space,
i.e. only closing *normal* prio and below.


rakshasa:
Merged. #1383

Issue #928, opened 2019-10-17
rtorrent (git) crashed due to 'save_input_history'


vp1981:
Hello,
I faced with rtorrent crash on Archlinux. The distribution packages `libtorrent` and `rtorrent` don't have debug symbols so I rebuild both packages from git repo. In gdb I see:
```
run
Starting program: /usr/bin/rtorrent 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[New Thread 0x7ffff6926700 (LWP 358966)]

Thread 1 "rtorrent" received signal SIGSEGV, Segmentation fault.
0x000055555565358e in ui::Root::save_input_history (this=0x55555576f930) at root.cc:462
462     root.cc: No such file or directory.
(gdb) bt
#0  0x000055555565358e in ui::Root::save_input_history (this=0x55555576f930) at root.cc:462
#1  0x00005555555b7e31 in std::__invoke_impl<void, void (core::DownloadList::*&)(), core::DownloadList*&> (__f=<optimized out>, __f=<optimized out>, __t=<optimized out>) at /usr/include/c++/9.2.0/bits/invoke.h:70
#2  std::__invoke<void (core::DownloadList::*&)(), core::DownloadList*&> (__fn=<optimized out>) at /usr/include/c++/9.2.0/bits/invoke.h:95
[...]

PR #977, opened 2020-03-25
Fix NULL pointer dereference

bacher09:
In some cases `m_control` could be `NULL` in that case rtorrent will crash. Here's minimal sample config that causes crash:

    session.save = yes

Also, wanted to say that this also affects **0.9.8** too.

### DIFF
--- a/src/ui/root.cc
+++ b/src/ui/root.cc
@@ -460,7 +460,7 @@ Root::load_input_history() {
 
 void
 Root::save_input_history() {
-  if (!m_control->core()->download_store()->is_enabled())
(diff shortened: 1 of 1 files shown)


vp1981:
Hello,
please see [issue 928](https://github.com/rakshasa/rtorrent/issues/928), may be you try to solve the same problem (it has PR, see last comment).

bacher09:
According to PR #929 it appears to be the same. Basically this bug is triggered by `session.save`.

bacher09:
Yeah, it looks like it's the same, when I opened I checked existing PR and none of them mention NULL pointer deference so I made this one.

PR #1095, opened 2021-07-04
Fixed a few memory leaks

stickz:
Backports a few memory leak fixes from [jesec/rtorrent](https://github.com/jesec/rtorrent) .

Requires the following [rakshasa/libtorrent](https://github.com/rakshasa/libtorrent) pull request to be merged first. https://github.com/rakshasa/libtorrent/pull/223

### DIFF
--- a/src/display/window_file_list.cc
+++ b/src/display/window_file_list.cc
@@ -62,14 +62,14 @@ WindowFileList::WindowFileList(const ui::ElementFileList* element) :
 // taking into account that some characters may be occupying two screen positions.
 std::wstring
 wstring_width(const std::string& i_str, int width) {
-  wchar_t result[width + 1];
-  size_t length = std::mbstowcs(result, i_str.c_str(), width);
+  wchar_t buf[width + 1];
+  size_t length = std::mbstowcs(buf, i_str.c_str(), width);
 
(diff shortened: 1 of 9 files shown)


samad909:
This PR (used with rakshasa/libtorrent#223) seems to be causing issues when SCGI is using a socket (network.scgi.open_local) instead of a port (network.scgi.open_port). rTorrent crashes with,
munmap_chunk(): invalid pointer
Aborted

src/display/window_file_list.cc
```diff
@@ -105,11 +105,15 @@ wstring_width(const std::string& i_str, int width) {
...
+    buf[length++] = ' ';
     swidth++;
   }
 
-  result[length] = 0;
+  buf[length] = 0;
+  
+  std::wstring result(buf);
+
+  free(buf);
```
TrimmingFool, line 116: @stickz On a cursory view, this makes no sense to me. Where do you see a memory leak? Moreover, `buf` is allocated on the  stack, freeing it is an error.

stickz:
I'm going to cherry pick these commits and create separate pull requests. This will allow @rakshasa to choose which ones to merge. Not all of these fixes are stable apparently. But they have a significant impact reducing software memory leaks.

PR #1098, opened 2021-07-19
Update README.md

ilyakmet:


rakshasa:
Thanks, however not going to merge. While I can see in the diff the addresses haven't been changed I think it's best to have a policy of never letting anyone except me touch the donation section.

PR #1131, opened 2021-11-22
Don't force the trackers to scrape when rTorrent first starts

stickz:
rTorrent will do this gradually, so we can start connecting to peers right away.

This patch shaves approximately 5 minutes off the start time with thousands of torrents.


rakshasa:
#1210 fixes this PR and was merged.

Issue #854, opened 2019-03-31
rtorrent crahes upon invalid XMLRPC logic


lps-rocks:
rTorrent can be crashed by sending invalid logic through XMLRPC interface. Issuing a d.multicall with an f. parameter causes rTorrent to crash unconditionally. XMLRPC commands should never cause a crash. They should be handled gracefully and an error returned to the connecting client.
using python xmlrpc.client library
```
import xmlrpc.client
server = xmlrpc.client.ServerProxy("http://127.0.0.1/rutorrent/plugins/rpc/rpc.php")
server.d.multicall("", "d.hash=", "d.is_multi_file=", "f.get_path=")
```
results in rTorrent crashing with
```
terminate called after throwing an instance of 'std::logic_error'
what():  basic_string::_M_construct null not valid
Aborted
```

kannibalox:
It turns out you could do the same crash with just `rtxmlrpc f.path <hash>`, looks like the target type check for files isn't actually using the template like the other targets.

PR #1251, opened 2023-10-01
Ensure compability check actually checks the type for files

kannibalox:
I'm actually not entirely sure if the other two checks are still necessary, but kept them in case there were some side effects I don't know about.

Fixes https://github.com/rakshasa/rtorrent/issues/854


kannibalox:
I have now grokked the two existing checks, and element_file_list.cc no longer throws an error.

rakshasa:
This should be `target.first == command_base::target_file || target.first == command_base::target_file_itr;`.

rakshasa:
Replaced by #1327.

Issue #490, opened 2016-08-19
config: Implement Lua-based configuration


cyphar:
Currently the configuration language is very stunted. Projects like NeoVim are introducing full programming language interpreters into things like editors and other tools. I reckon it would be pretty cool if we could do a similar thing for `rtorrent`. This would remove the need for the currently fairly ugly method of running commands with events (everything could be done with callbacks or something, depending on what API you want to expose). I would recommend using LuaJIT as the base (C implementation, very fast) and adding hooks to the relevant C++ code in the evaluator.

The reason for wanting this is that the `rtorrent` configuration language is clearly a good usecase for a proper language (it is already trying to be one with hook support, but it's clearly not designed to be a language from the outset).

WDYT

cyphar:
This would also make things like #293 much easier to support -- you just have to add more bindings to the Lua interpreter so that it can modify more internal state of rtorrent. There is a valid question however about what will happen with the RPC interpreter for rtorrent (since you could implement it in Lua, but understandably that's probably something you might not want -- though implementing it in Lua would make running Lua callbacks based on RPC commands possible).

pyroscope:
So if Jari acccepts this in principle, who's actually able to work on it? I'd pitch in, but won't / cannot do it alone.

rakshasa:
I'm ok with this, essentially the current syntax is just a parser used to call commands in rtorrent.

So we currently have the parser used for e.g. rtorrent.rc, etc, and then we have the XMLRPC wrapper. Adding another one should not be that difficult.

pyroscope:
IMHO this should be done in small increments, and providing a clear migration path, so this would be a roadmap proposal (each step should be its own ticket pointing to this story):

### Step 1: Add LuaJIT to the project

* Add LuaJIT to the project, as an optional dependency, using a ``--with-lua-jit`` option.
* Add a new ``lua.*`` command family similar to ``execute.*``, as a test bed and to allow people to slowly grow into Lua with their existing configs. This would solve #476.

### Step 2: Callback integration

* Add callbacks from Lua into rtorrent (call commands as functions).
* Possibly offer value-type commands as attributes of an rtorrent object or similar.

### Step 3: Lua as the configuration engine

* Look for a ~/.rtorrent.lua file in addition to the existing config.
* Modify import to recognize ``*.lua`` files.


:tada: Done! :checkered_flag:

cyphar:
[...]

PR #1260, opened 2024-01-18
Add optional Lua scripting capabilties

kannibalox:
This was tested on Lua 5.4, since while there shouldn't be any version-specific features in use, the lack of native 64-bit integer support in versions before 5.3 seems like a deal breaker.

There are three RPC commands added:
`lua.execute`, which allows executing Lua code, with a `rtorrent` package that has basic support for making direct RPC calls from a file:
```lua
local ofile = io.open("/tmp/out.lua.txt", "w")
-- Purposefully inefficient as a demonstration of `rtorrent.call`
for _, result in ipairs(rtorrent.call("d.multicall2", "", "main", "d.hash=")) do
   ofile:write(rtorrent.call("d.name", result[1]).."\n")
end
-- Alternative syntactical sugar
r = rtorrent.autocall
ofile:write(r.system.hostname())
ofile:close()
```

That's a little verbose to be a replacement for `rtorrent.rc`, so there's also `lua.import`, which uses the same "autocall" Lua magic for a simpler syntax:
```lua
local base = "/tmp/rtorrent/"
scgi_port("0.0.0.0:9998")
session(base.."session")
directory(base.."data")
network.http.max_open.set(256)
```

`lua.execute.str` is mainly to allow events, but could it definitely use a more convenient syntax:
```lua
local function insert_method(name, func_name, arg0, arg1, arg2, arg3)
   args = {
      arg0 or "$argument.0=",
      arg1 or "$argument.1=",
      arg2 or "$argument.2=",
      arg3 or "$argument.3=",
   }
   arg_str = table.concat(args, ",")
   rtorrent.call("method.insert", "", name, "simple", "lua.execute.str=\"return "..func_name.."(...)\","..arg_str)
end
function event_handler(target, name)
   rtorrent.call("print", "", "Lua event: Inserted \""..target..": "..name.."\"")
end
insert_method("d.event_handler", "event_handler", "(d.name)")

rtorrent.call("method.set_key", "", "event.download.inserted", "print_name", "d.event_handler=")
```

Implements https://github.com/rakshasa/rtorrent/issues/490

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -56,6 +56,8 @@ AC_LANG_PUSH(C++)
 TORRENT_WITH_XMLRPC_C
 AC_LANG_POP(C++)
 
+TORRENT_WITH_LUA
+
 AC_DEFINE(HAVE_CONFIG_H, 1, true if config.h was included)
 AC_DEFINE(USER_AGENT, [std::string(PACKAGE "/" VERSION "/") + torrent::version()], Http user agent)
 
--- a/rtorrent.lua
+++ b/rtorrent.lua
@@ -1,0 +1,42 @@
+-- the "rtorrent" table is passed in by the C++ code, modify and
+-- return it for loading.
+local args = {...}
+local rtorrent = args[1]
+
+-- Autocall
+-- Allows syntax like `rtorrent.autocall.system.hostname()`
+local mt = {}
+function mt.__call (t, ...)
+   name = table.concat(rawget(t, "__namestack"), ".")
+   success, ret = pcall(rtorrent.call, name, ...)
+   if not success then error(name..": "..ret, 2) end
+   return ret
+end
+function mt.__index (t, key)
+   ns = rawget(t, "__namestack") or {}
+   table.insert(ns, key)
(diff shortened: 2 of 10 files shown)


kartikynwa:
This looks interesting. What do you use this for personally?

kannibalox:
I converted my config to Lua, but that was more to test any edge cases. More regularly I find myself using it to do complicated client-side filtering.

My main goal was to start out with making methods/events more structured, since that proves to be a major source of confusion when new people are learning the rTorrent scripting language. Currently I think it needs a better syntax than the one I provided above, so I'm still playing around with that a little. If you're looking for ideas, [luarocks](https://luarocks.org/) has a lot of neat packages that could extend rtorrent to do pretty much anything.

rakshasa:
Merged #1380.

Git-Boy-Slim:
Now that this feature is merged could you maybe add a little more information about it or some examples in the documentation @kannibalox?

Can this completely replace the scripting language of rtorrent? 

This looks very promising, but I can't wrap my head around how I could translate my existing config.rc files into lua.

PR #1293, opened 2024-09-06
Fix session file data corruption

stickz:
From https://github.com/jesec/rtorrent/commit/f9a875b633eb083ec82f89ebeae8678ca424e8b2

This commit ensures session files land on the disk in the event of a system crash or power outage. It prevents data corruption of rTorrent session files, which causes torrent client breakage.

Optimizations from the initial implementation include:
1. Skipping unnecessary file metadata updates during the sync process. 
2. Skipping variable initialization since open will assign -1 on failure.

### DIFF
--- a/src/core/download_store.cc
+++ b/src/core/download_store.cc
@@ -40,6 +40,7 @@
 
 #include <fstream>
 #include <stdio.h>
+#include <fcntl.h>
 #include <unistd.h>
 #include <rak/error_number.h>
 #include <rak/path.h>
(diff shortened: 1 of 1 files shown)


jesec:
I would recommend using fsync, since fdatasync is not well defined and its behavior is not consistent across different OSs and file systems. 

https://news.ycombinator.com/item?id=25171572

By the way, next time try using “git cherry-pick” or “git am” to pick commits.

rakshasa:
I'm not going to make the default slower because some OSs might not do the size metadata sync.

If this is an issue, it should be addressed by detecting those OSs in configure.

PR #1296, opened 2024-09-29
Fix static building

stickz:
@rakshasa A major hosting provider with rTorrent has mentioned the following revision is required for static builds to pass.


eli-schwartz:
This looks unnecessary to me. rtorrent is a program, not a libtool library. The commit adds LT_INIT and invokes the libtool machinery to compile and link the /usr/bin/rtorrent binary instead of directly running the compiler -- dependencies are anyways fetched via pkg-config.

I don't see how using LT_INIT can fix static building. The resulting `./configure --enable-static --enable-shared` options aren't even utilized for building binaries, only for building libraries, which would matter for libtorrent, not rtorrent.

It's a bit slower to build using the libtool proxy and I don't see any benefit to using it here. And if I wanted a static build of rtorrent, I would simply build libtorrent with `--enable-static --disable-shared` and then compile rtorrent with LDFLAGS="-static".

Am I missing something here?

stickz:
> 
> Am I missing something here?
Yeah, a hosting provider was creating a static build of rtorrent using musl. This resolved their problem. All dependencies including the c standard library were static.

PR #1297, opened 2024-09-29
fix: use `fsync` for osx builds

chenrui333:
relates to https://github.com/Homebrew/homebrew-core/pull/192310


chenrui333:
cc @rakshasa

PR #1307, opened 2024-10-17
Simplify updating queue entries

kannibalox:
The erase/insert pattern can be replaced by a single method that updates an existing entry's timer, or inserting if it doesn't exist.

This has much less impact than the libtorrent change it's based on, but might as well be consistent.


stickz:
@rakshasa This change is stable and has merit. Could we get a review?

Issue #1250, opened 2023-09-26
`f.path=` return invalide string name if it contain emoji


trim21:
I have a torrent (I can provide if you need) which all file paths are utf-8 encoded. But becoming invalid string in rpc call:

for example:  
`套图/推特/211031_万圣节快乐🎃不给🍭不给看🧛‍♀️🧛‍♀️🧛‍♀️.jpg` become `??????/??????/211031_?????????????????????????????????????????????????????????????????????????????.jpg`

This will be fixed by https://github.com/rakshasa/rtorrent/pull/1309 , you will need to compile rtorrent with `--with-xmlrpc-tinyxml2`

But it's not released yet, may be included in rtorrent>0.10.0

kannibalox:
Unfortunately this is a limitation of the xmlrpc-c library, which only allows unicode characters from the basic multilingual plane: https://xmlrpc-c.sourceforge.io/doc/libxmlrpc_util.html#validate_utf8 (emoji are from the supplementary plane)

trim21:
> Unfortunately this is a limitation of the xmlrpc-c library, which only allows unicode characters from the basic multilingual plane: https://xmlrpc-c.sourceforge.io/doc/libxmlrpc_util.html#validate_utf8 (emoji are from the supplementary plane)

said 😢 

I found that https://github.com/jesec/rtorrent support json rpc and doesn't have this issue.

trim21:
It's fixed now

PR #1309, opened 2024-10-20
Allow using vendored tinyxml2 for XMLRPC

kannibalox:
This removes the dependency on xmlrpc-c in favor of embedding a XML library ([tinyxml2](http://leethomason.github.io/tinyxml2/)) and doing the parsing ourselves, heavily inspired by https://github.com/jesec/rtorrent/tree/refactor/xml-as-json. Many of the things xmlrpc-c would normally handle rTorrent already does, e.g. centrally mapping method names to functions is already done by CommandMap.

This generally makes compiling easier (tinyxml2 compiles across a wide variety of systems), and fixes limitations of xmlrpc-c, such as mangling emojis as per https://github.com/rakshasa/rtorrent/issues/1250.

I did not re-implement all the [introspection methods](https://xmlrpc-c.sourceforge.io/doc/libxmlrpc_server.html#systemmethod), only `system.listMethods` and `system.multicall`, but the rest could be added if needed. I also intentionally made the network.xmlrpc.* commands noops, since there's not much need to manage dialect, and it doesn't make sense to try and control memory usage at the XML level (it should happen at the SCGI level). The main downside is that a bunch of error checking had to be re-implemented. tinyxml2 also only officially supports UTF-8.

The choice of tinyxml2 over another library like rapidxml/pugixml mostly stems from it's ability to build output as a stream rather than having to specifically build an entire DOM and print it all at once. rTorrent typically spends more time returning a response than it does parsing the request, so it seemed like a better tradeoff. However, all a library would need to be able to do is parse and print XML and it could be swapped in, in case there's some pressing reason to use a different one.


scripts/checks.m4
```diff
@@ -384,49 +384,6 @@ AC_DEFUN([TORRENT_WITH_FASTCGI], [
...
-        AC_MSG_ERROR(Could not compile XMLRPC-C test.)
-      fi
-    fi
-
-  ],[
-    AC_MSG_RESULT(ignored)
-  ])
-])
-
-
```
stickz, line 429: Could we still keep `xmlrpc-c` as an option? Then add `tinyxml2` as an alternative option? At least ten thousand users are running `ruTorrent` with `xmlrpc-c` currently. We need time to test this before dropping `xmlrpc-c`.

src/Makefile.am
```diff
@@ -188,6 +188,10 @@ libsub_root_a_SOURCES = \
 	thread_base.cc \
 	thread_base.h \
 	thread_worker.cc \
-	thread_worker.h
+	thread_worker.h \
+	\
```
stickz, line 192: Syntax error.

rakshasa:
I'm for this change, as I like to limit bloated external dependencies if there's simple implementations we can include directly.

However this would need to be implemented in stages, with xmlrpc-c as default first for a couple of releases, then default to tinyxml2 then remove the former completely in a couple years.

src/Makefile.am
kannibalox: Can you describe the syntax error for me? I'm unable to reproduce on a fresh checkout, and the same pattern is used farther up in the file.

kannibalox:
xmlrpc-c isn't part of the default build (i.e. if you just do a raw `./configure` it won't be included), are you saying you'd like me to change it to be? It seems like it'd be easier to allow tinyxml2 to be the default, and then if `--with-xmlrpc-c` is provided (as it is in nearly all existing build scripts) that would override it. Alternatively I can keep both off by default and add a new flag called `--with-tinyxml2`.

src/Makefile.am
stickz: We don't need two forward slashes in a row. The compiler is just going to the next line twice. It won't throw an error, but should still fix it.

stickz:
> xmlrpc-c isn't part of the default build (i.e. if you just do a raw `./configure` it won't be included), are you saying you'd prefer it to be? It seems like it'd be easier to allow tinyxml2 to be the default, and then if `--with-xmlrpc-c` is provided (as it is in nearly all existing build scripts) that would override it. Alternatively I can keep both off by default and add a new flag called `--with-tinyxml2`.

We need `--with-tinyxml2` first. We can't enable `tinyxml2` by default until it's tested. 

After a couple releases, we can enable it by default and override it if `--with-xmlrpc-c` is specified. But we should still add `--without-tinyxml2` to give the option to exclude. This would allow us to create a minimal build of the torrent client.

kannibalox:
I'd argue that tinyxml2 is minimal, and if people are actually using the minimal build already then they aren't using the XMLRPC part at all, but will defer to @rakshasa's decision.

I'm adding some unit tests for correctness, is there anything else you'd like to see tested? Few people have access to 10k ruTorrent clients for testing, if there's any way I can increase confidence in-code that'd be good.

src/Makefile.am
kannibalox: I'm still confused as to the issue, I'm simply following the same pattern as above in the same file, where an extra line and forward slash is used to separate directories. Is it a syntax error, just a warning, or something else?

scripts/checks.m4
kannibalox: Re-added xmlrpc-c support, the exact details of what should be enabled by default is being discussed in the main comments

kannibalox:
I've re-added xmlrpc-c support, simple unit testing (which is disabled for xmlrpc-c because it has just a slightly different output format, and I'm too lazy to try and make it work for a dependency that I'm trying to get rid of), and changed a couple things to be more inline with how xmlrpc-c operates.

kannibalox:
Here are some rough benchmarks for the performance side of things:
xmlrpc-c:
```
Benchmark               Time             CPU   Iterations
---------------------------------------------------------
small_response       3417 ns         3413 ns       204332
large_response   46159260 ns     46074301 ns           15
```
tinyxml2:
```
Benchmark               Time             CPU   Iterations
---------------------------------------------------------
small_response       1595 ns         1593 ns       462388
large_response    1512614 ns      1509299 ns          463
```
Benchmark source, using https://github.com/google/benchmark
<details>

```cpp
#include <benchmark/benchmark.h>

#include "config.h"
#include <string>
#include <iostream>

#include "command_helpers.h"
#include "rpc/command_map.h"

#include "control.h"
#include "globals.h"

torrent::Object g_obj;
torrent::Object xmlprpc_cmd_test_map_a(rpc::target_type t, const torrent::Object& obj) { return g_obj; }
torrent::Object xmlprpc_cmd_test_map_b(rpc::target_type t, const torrent::Object& obj) { return obj; }

static void DoSetup(const benchmark::State& state) {
  std::cout << "setup\n";
  control = new Control;
  setlocale(LC_ALL, "");
  cachedTime = rak::timer::current();
  if (!rpc::xmlrpc.is_valid())
    rpc::xmlrpc.initialize();
  g_obj = torrent::Object::create_list();
  for (int i=0;i<10000; i++) {
    auto submap = torrent::Object::create_map();
    submap.as_map()["hash"] = 'A'*40;
    submap.as_map()["size"] = 12414124141;
    submap.as_map()["size2"] = 12414124141;
    g_obj.as_list().push_back(submap);
  }
  if (rpc::commands.find("xmlrpc_reflect") == rpc::commands.end()) {
    CMD2_ANY("xmlrpc_reflect", &xmlprpc_cmd_test_map_a);
    CMD2_ANY("xmlrpc_reflect2", &xmlprpc_cmd_test_map_b);
  }
}


static void small_response(benchmark::State& state) {
  std::string input("<?xml version=\"1.0\"?><methodCall><methodName>xmlrpc_reflect2</methodName><params></params></methodCall>");
  for (auto _ : state) {
    rpc::xmlrpc.process(input.c_str(), input.size(), [](const char*, uint32_t){return true;});
  }
}

static void large_response(benchmark::State& state) {
  std::string input("<?xml version=\"1.0\"?><methodCall><methodName>xmlrpc_reflect</methodName><params></params></methodCall>");
  for (auto _ : state) {
    rpc::xmlrpc.process(input.c_str(), input.size(), [](const char*, uint32_t){return true;});
  }
}
// Register the function as a benchmark
BENCHMARK(small_response)->Setup(DoSetup);
BENCHMARK(large_response);
// Run the benchmark
BENCHMARK_MAIN();
```
</details>

kannibalox:
Done. If both flags are set, xmlrpc-c takes precedence.

stickz (approved):

src/rpc/xmlrpc.cc
```diff
@@ -36,12 +36,18 @@
...
+#include <string>
+#include <sstream>
+
+#if defined(HAVE_XMLRPC_C)
 #include <stdlib.h>
 #include <xmlrpc-c/server.h>
+#else
+#include "vendor/tinyxml2/tinyxml2.h"
+#include "utils/base64.h"
 #endif
```
rakshasa, line 49: Should check if we have tinyxml.

src/rpc/xmlrpc.cc
```diff
@@ -607,6 +613,357 @@ XmlRpc::set_size_limit(uint64_t size) {
   xmlrpc_limit_set(XMLRPC_XML_SIZE_LIMIT_ID, size);
 }
 
+bool
+XmlRpc::is_valid() const { return m_env != NULL; }
+
+#elif defined(HAVE_XMLRPC_TINYXML2)
```
rakshasa, line 619: Put these in separate files, `xmlrpc_xmlrpc_c.cc`, `xmlrpc_tinyxml2.cc`, etc.

src/rpc/xmlrpc.cc
```diff
@@ -607,6 +613,357 @@ XmlRpc::set_size_limit(uint64_t size) {
...
+};
+
+const tinyxml2::XMLElement* element_access(const tinyxml2::XMLElement* elem, std::string elemNames) {
+  // Helper function to check each step of a element access, in liueu of XPath
+  char token = ',';
+  std::stringstream ss(elemNames);
+  std::string item;
+  std::string stack("");
+  const tinyxml2::XMLElement* result = elem;
+  while (getline (ss, item, token)) {
```
rakshasa, line 641: Using stringstream is inefficient, especially when you know you're searching for a single-char token.

Iterators and std::find should be enough.

src/rpc/xmlrpc.cc
kannibalox: Will be handled by the other request to split out into different files

src/rpc/xmlrpc.cc
kannibalox: Done, now it uses the config.h flags to enable/disable code, and resolution of which one to use is more of a link time decision. It wasn't my first choice, I did try to go with a OOP model, but that would've leaked changes into other places, including the existing xmlrpc-c code, which seemed to be worth avoiding.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+//
+// Contact:  Jari Sundell <jaris@ifi.uio.no>
+//
+//           Skomakerveien 33
+//           3185 Skoppum, NORWAY
+
+#include "config.h"
+
+#ifdef HAVE_XMLRPC_TINYXML2
+#ifndef HAVE_XMLRPC_C
```
rakshasa, line 40: Having both defined should result in an `#error` in both files.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+private:
+  int                 m_type;
+  std::string         m_msg;
+};
+
+const tinyxml2::XMLElement*
+element_access(const tinyxml2::XMLElement* elem, std::string elemNames) {
+  // Helper function to check each step of a element access, in lieu of XPath
+  char token = ',';
+  std::string stack("");
```
rakshasa, line 88: No need for `token` and don't use `("")`.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+  const tinyxml2::XMLElement* result = elem;
+  size_t start = 0;
+  size_t end = 0;
```
rakshasa, line 91: Rewrite to only use `size_t pos = 0;`.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+  do {
+    end = elemNames.find(token, start);
+    auto item = elemNames.substr(start, end - start);
+    stack.append(item);
+    result = result->FirstChildElement(item.c_str());
+    if (result == nullptr) {
+      throw xmlrpc_error(XMLRPC_PARSE_ERROR, "could not find expected element " + stack);
+    }
+    stack.append("->");
```
rakshasa, line 100: Is the `stack` only there for errors?

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+  } while (end != std::string::npos);
+  return result;
+}
+
+torrent::Object
+xml_value_to_object(const tinyxml2::XMLNode* elem) {
+  if (elem == nullptr) {
+    throw xmlrpc_error(XMLRPC_INTERNAL_ERROR, "received null element to convert");
+  }
+  if (std::string("value") != elem->Value()) {
```
rakshasa, line 111: Use `std::strncmp`.

src/rpc/xmlrpc_tinyxml2.cc
kannibalox: Correct, I originally added it to help myself out troubleshooting but thought it might be useful in general. It can be removed if you'd prefer.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+    throw xmlrpc_error(XMLRPC_INTERNAL_ERROR, "received non-value element to convert");
+  }
+  auto valueElem = elem->FirstChild();
+  auto elemType = std::string(valueElem->Value());
```
rakshasa, line 115: `element_value`, `element_type`, etc

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+  if (elemType == "string") {
+    auto elemChild = valueElem->FirstChild();
+    if (elemChild == nullptr) {
+      return torrent::Object("");
+    }
+    return torrent::Object(elemChild->ToText()->Value());
+  } else if (elemType == "i4" or elemType == "i8" or elemType == "int") {
```
rakshasa, line 122: use `||`

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+    auto elemValue = std::string(valueElem->FirstChild()->ToText()->Value());
+    return torrent::Object(std::stol(elemValue));
```
rakshasa, line 124: `std::strtol` with no intermediate, 

```
char* pos;
auto str = valueElem->FirstChild()->ToText()->Value();
auto result = std::strtoll(str, &pos);

if (pos == str || *pos != '\0')
  // error, nothing read or partial read

return ...
```

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+  } else if (elemType == "boolean") {
+    auto boolText = std::string(valueElem->FirstChild()->ToText()->Value());
+    if (boolText == "1") {
+      return torrent::Object((int64_t)1);
+    } else if (boolText == "0") {
+      return torrent::Object((int64_t)0);
+    }
```
rakshasa, line 131: Error if neither.

src/rpc/xmlrpc_tinyxml2.cc
kannibalox: FWIW, did a quick benchmark and it optimizes out the same. Changed for consistent style though.

src/rpc/xmlrpc_tinyxml2.cc
rakshasa: Don't do extraneous work just for errors, if absolutely needed it should be redone in the error branch.

src/rpc/xmlrpc_tinyxml2.cc
rakshasa: It's not a benchmark thing, it's for code clarity and proper location of variables inside/outside the while block.

src/Makefile.am
```diff
@@ -188,6 +190,10 @@ libsub_root_a_SOURCES = \
...
+	vendor/tinyxml2/tinyxml2.h \
+	vendor/tinyxml2/tinyxml2.cpp
```
rakshasa, line 196: Don't like using vendor directory here, better to use `rpc/tinyxml2/tinyxml2.h` and `namespace rpc::tinyxml2` imo.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+
+const tinyxml2::XMLElement*
+element_access(const tinyxml2::XMLElement* elem, std::string element_names) {
+  // Helper function to check each step of a element access, in lieu of XPath
+  const tinyxml2::XMLElement* result = elem;
+  size_t pos = 0;
+  do {
+    auto previous_pos = pos;
+    pos = element_names.find(',', pos + 1);
+    auto item = element_names.substr(previous_pos, pos - previous_pos);
```
rakshasa, line 94: You cannot use the `npos` case to calculate the size here.

```
while (pos < element_names.size()) {
  auto prev = pos;
  pos = element_names.find(',', pos);

  if (pos == std::string::npos)
    pos = element_names.size();

  if (pos == prev) {
    // handle empty, should be error?
    pos++;
    continue;
  }

  auto item = element_names.substr(prev, pos - prev);
  pos++;
```

This would be cleaner if you used iterators instead of pos.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+    } else if (boolean_text == "0") {
+      return torrent::Object((int64_t)0);
+    }
+    throw xmlrpc_error(XMLRPC_TYPE_ERROR, "unknown boolean value: " + boolean_text);
+  } else if (value_element_type == "array") {
+    auto array = torrent::Object::create_list();
+    auto data_element = element_access(value_element->ToElement(), "data");
+    for (auto array_value_element = data_element->FirstChildElement("value"); array_value_element; array_value_element = array_value_element->NextSiblingElement("value")) {
+      array.as_list().push_back(xml_value_to_object(array_value_element));
+    }
```
rakshasa, line 139: Write this as a while loop.

rakshasa: Same below, the for loops are too messy to read.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+    auto value_element_child = value_element->FirstChild();
+    if (value_element_child == nullptr) {
+      return torrent::Object("");
+    }
+    auto base64string = std::string(value_element_child->ToText()->Value());
+    base64string.erase(std::remove_if(base64string.begin(), base64string.end(),
+                                      [](char c) { return c == '\n' || c == '\r'; }),
+                       base64string.end());
+    return torrent::Object(utils::base64decode(base64string));
+    return torrent::Object("");
```
rakshasa, line 158: Remove.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+  size_t pos = 0;
+  do {
+    auto previous_pos = pos;
+    pos = element_names.find(',', pos + 1);
+    auto item = element_names.substr(previous_pos, pos - previous_pos);
+    result = result->FirstChildElement(item.c_str());
+    if (result == nullptr) {
+      throw xmlrpc_error(XMLRPC_PARSE_ERROR, "could not find expected element " + item);
+    }
+  } while (pos != std::string::npos);
```
rakshasa, line 99: This would be cleaner if you used iterators instead of pos.

You're implicitly assuming there is no `,,` in the string by doing `pos + 1`.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,454 @@
...
+  CommandMap::iterator cmd_itr = commands.find(method_name.c_str());
+  if (cmd_itr == commands.end() || !(cmd_itr->second.m_flags & CommandMap::flag_public_xmlrpc)) {
+    throw xmlrpc_error(XMLRPC_NO_SUCH_METHOD_ERROR, "Method '" + std::string(method_name) + "' not defined");
+  }
+  torrent::Object params_raw = torrent::Object::create_list();
+  torrent::Object::list_type& params = params_raw.as_list();
+  if (params_element != nullptr) {
+    for (auto param_element = params_element->FirstChildElement("param"); param_element; param_element = param_element->NextSiblingElement("param")) {
+      params.push_back(xml_value_to_object(param_element->FirstChildElement("value")));
+    }
```
rakshasa, line 315: Not liking overloading of names here, just use `child` or something.

src/rpc/xmlrpc_tinyxml2.cc
kannibalox: I'm thinking this is no longer worth being a separate function, and should just be unrolled in the cases where I'm using it with one field, and using simple initializer syntax in the cases where it still helps to remove boilerplate, e.g. `auto value_elements = element_access(doc->RootElement(), "params,param,value,array,data");` becomes
```cpp
for (auto itr : {"params", "param", "value", "array", "data"}) {
  child = result->FirstChildElement(itr);
  if (child == nullptr)
     throw xmlrpc_error(XMLRPC_PARSE_ERROR, "could not find expected element " + itr)
}
```
That way the CSV-ness is enforce by the syntax, and the list is 100% static. I also thought about passing it as a std::initializer list to the function, but it seems like there's some [hidden copying](https://akrzemi1.wordpress.com/2016/07/07/the-cost-of-stdinitializer_list/) there.
Thoughts?

kannibalox: Went ahead with a hybrid approach for now: unrolled the single accesses, and changed it to use an initializer list.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,451 @@
...
+    break;
+  default:
+    printer->OpenElement("i4", true);
+    printer->PushText(0);
+    printer->CloseElement(true);
+  }
+}
+
+void
+object_to_target(const torrent::Object& obj, int callFlags, rpc::target_type* target) {
```
rakshasa, line 231: This should also be used for xmlrpc-c's `xmlrpc_to_target`.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,451 @@
...
+    throw torrent::input_error("invalid parameters: invalid index");
+  }
+}
+
+torrent::Object execute_command(std::string method_name, const tinyxml2::XMLElement* params_element) {
+  if (params_element == nullptr)
+    throw xmlrpc_error(XMLRPC_INTERNAL_ERROR, "invalid parameters: null");
+  CommandMap::iterator cmd_itr = commands.find(method_name.c_str());
+  if (cmd_itr == commands.end() || !(cmd_itr->second.m_flags & CommandMap::flag_public_xmlrpc)) {
+    throw xmlrpc_error(XMLRPC_NO_SUCH_METHOD_ERROR, "Method '" + std::string(method_name) + "' not defined");
```
rakshasa, line 305: `method_name` is already a std::string

rakshasa: Lower-case `Method`

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,451 @@
...
+      params.push_back(xml_value_to_object(child->FirstChildElement("value")));
+    }
+  }
+  rpc::target_type target = rpc::make_target();
+  if (params.size() == 0 && (cmd_itr->second.m_flags & (CommandMap::flag_file_target | CommandMap::flag_tracker_target))) {
+    throw xmlrpc_error(XMLRPC_TYPE_ERROR, "invalid parameters: too few");
+  }
+  if (params.size() > 0) {
+    object_to_target(params.front(), cmd_itr->second.m_flags, &target);
+    params.erase(params.begin());
```
rakshasa, line 320: Instead of erasing the target param here, just store it separately from the start.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,451 @@
...
+  if (doc->FirstChildElement("methodCall")->FirstChildElement("methodName") == nullptr)
+    throw xmlrpc_error(XMLRPC_PARSE_ERROR, "methodName element not found");
+  auto method_name = doc->FirstChildElement("methodCall")->FirstChildElement("methodName")->GetText();
+  torrent::Object result;
+
+  // Add a shim here for system.multicall to allow better code reuse, and
+  // because system.multicall is one of the few methods that doesn't take a target
+  if (method_name == std::string("system.multicall")) {
+    result = torrent::Object::create_list();
+    torrent::Object::list_type& result_list = result.as_list();
```
rakshasa, line 340: Can just use `auto&`.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,451 @@
...
+}
+
+void XmlRpc::initialize() { m_isValid = true; }
+void XmlRpc::cleanup() {}
+
+void XmlRpc::insert_command(const char*, const char*, const char*) {}
+void XmlRpc::set_dialect(int) {}
+
+int64_t XmlRpc::size_limit() { return std::numeric_limits<int64_t>::max(); }
+void    XmlRpc::set_size_limit(uint64_t) {}
```
rakshasa, line 445: Size limit needs to be a sane value, and in the same orders of magnitude that xmlrpc supports. (

Also, setting the size limit should have an effect, and be checked in `process(...)`.

Scgi sets `max_content_size    = (2 << 23);`.

src/utils/base64.h
```diff
@@ -0,0 +1,59 @@
+#include <string>
+
+#include <torrent/exceptions.h>
+
+namespace utils {
+constexpr static char padCharacter = '=';
+static std::string
+base64decode(const std::string& input) {
```
rakshasa, line 8: Where is this from?

src/rpc/xmlrpc_tinyxml2.cc
kannibalox: I did see that comment, but had deferred on the side of caution since the PR aims to replace that code entirely and there were already concerns about stability.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,451 @@
...
+  if (elem == nullptr) {
+    throw xmlrpc_error(XMLRPC_INTERNAL_ERROR, "received null element to convert");
+  }
+  if (std::strncmp(elem->Value(), "value", 5) != 0) {
+    throw xmlrpc_error(XMLRPC_INTERNAL_ERROR, "received non-value element to convert");
+  }
+  auto value_element = elem->FirstChild();
+  auto value_element_type = std::string(value_element->Value());
+  if (value_element_type == "string") {
+    auto value_element_child = value_element->FirstChild();
```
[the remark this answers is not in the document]
kannibalox: I mainly wanted to avoid confusion with the doc->RootElement(), and was intended to reflect the `<value>X</value>` element that was passed in, will change though.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,451 @@
...
+    if (value_element_child == nullptr) {
+      return torrent::Object("");
+    }
+    return torrent::Object(value_element_child->ToText()->Value());
+  } else if (value_element_type == "i4" || value_element_type == "i8" || value_element_type == "int") {
+    char* pos;
+    auto str = value_element->FirstChild()->ToText()->Value();
```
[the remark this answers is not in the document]
kannibalox: It is not, fixed and added a test case

src/rpc/tinyxml2/tinyxml2.cpp
```diff
@@ -0,0 +1,3031 @@
+// rtorrent-specific check
+#include "config.h"
+#ifdef HAVE_XMLRPC_TINYXML2
+/*
+Original code by Lee Thomason (www.grinninglizard.com)
```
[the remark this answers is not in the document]
kannibalox: Needed to be a bit lower to exclude the headers, but done

src/utils/base64.h
kannibalox: I originally stole it from  https://github.com/jesec/rtorrent/blob/199e8f85244c8eb1c30163d51755570ad86139bb/src/core/manager.cc#L46, but it looks to be a modified version of https://en.wikibooks.org/wiki/Algorithm_Implementation/Miscellaneous/Base64#C++_2

src/rpc/xmlrpc_tinyxml2.cc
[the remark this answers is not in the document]
rakshasa: Parent might be an alternative.

This is the difference between naming it what it is vs. what role it plays. E.g. calling someone a parent is often more descriptive than calling them a human.

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,459 @@
...
+  }
+  auto value_element = elem->FirstChild();
+  auto value_element_type = value_element->Value();
+  if (std::strncmp(value_element_type, "string", sizeof("string")) == 0) {
+    auto value_element_child = value_element->FirstChild();
+    if (value_element_child == nullptr) {
+      return torrent::Object("");
+    }
+    return torrent::Object(value_element_child->ToText()->Value());
+  } else if (std::strncmp(value_element_type, "int", sizeof("int")) == 0 || std::strncmp(value_element_type, "i4", sizeof("i4")) == 0|| std::strncmp(value_element_type, "i8", sizeof("i8")) == 0) {
```
rakshasa, line 129: multiple lines

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,459 @@
...
+      throw xmlrpc_error(XMLRPC_PARSE_ERROR, "could not find expected data element in array");
+    for (auto child = data_element->FirstChildElement("value"); child; child = child->NextSiblingElement("value")) {
+      array.as_list().push_back(xml_value_to_object(child));
+    }
+    return array;
+  } else if (std::strncmp(value_element_type, "struct", sizeof("struct")) == 0) {
+    auto map = torrent::Object::create_map();
+    for (auto child = value_element->FirstChildElement("member"); child; child = child->NextSiblingElement("member")) {
+      auto key = child->FirstChildElement("name")->GetText();
+      map.as_map()[key] = xml_value_to_object(child->FirstChildElement("value"));
```
rakshasa, line 152: all `as_*` should be done outside the loop

src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -0,0 +1,459 @@
...
+  } else if (std::strncmp(value_element_type, "base64", sizeof("base64")) == 0) {
+    auto value_element_child = value_element->FirstChild();
+    if (value_element_child == nullptr) {
+      return torrent::Object("");
+    }
+    auto base64string = std::string(value_element_child->ToText()->Value());
+    base64string.erase(std::remove_if(base64string.begin(), base64string.end(),
+                                      [](char c) { return c == '\n' || c == '\r'; }),
+                       base64string.end());
+    return torrent::Object(utils::base64decode(base64string));
```
rakshasa, line 164: Rename to `utils::decode_base64` and add a `utils::remove_newlines`.

src/rpc/xmlrpc_tinyxml2.cc
rakshasa: Better to get it done now that it's being worked on, rather than left as a todo that never gets done.

This will also test the code with the xmlrpc-c users.

src/utils/base64.h
rakshasa: Add the reference and that it is public domain as a comment.

Rename `padCharacter` to `base64_pad_character`.

Remove 'static' from the function definition.

Also, add newlines between functions, namespaces, etc.

src/rpc/xmlrpc_tinyxml2.cc
kannibalox: I genuinely can't imagine a situation where anyone would desire it to be anything other than the exact SCGI max length (no point is going higher, and content size is no predictor of CPU/memory usage, so going lower doesn't help even in low-spec situations), so I've had the default reference that value directly.

xmlrpc-c uses size_t for its limit, but I don't see a compelling reason to limit tinyxml2 to that, especially when it's capped by a hard-coded SCGI max content size in both cases anyway.

src/rpc/xmlrpc_tinyxml2.cc
kannibalox: The capital `Method` is actually because that's exactly how `xmlrpc-c` responds. I've changed it already, but wanted to bring it to your attention.

rakshasa:
Except for some minor style issues, it's looking mergable.

Probably going to add some CI tests to rtorrent-docker for xmlrpc requests first, so might take a few days.

rakshasa:
Merged.

trim21:
I saw #1250 is mentioned, doesn't this mean tinyxml2 support utf-8?

rakshasa:
Afaik tinyxml2 doesn't do anything to the strings, so it should work.

trim21:
looks like its still broken for emoji support.

response for filename `🤨`

b'<?xml version="1.0" encoding="UTF-8"?>\r\n<methodResponse>\r\n<params>\r\n<param><value><array><data>\r\n<value><array><data>\r\n<value><string>????</string></value>\r\n<value><i8>6114656256</i8></value>\r\n</data></array></value>\r\n</data></array></value></param>\r\n</params>\r\n</methodResponse>\r\n'

trim21:
Oh, sorry, maybe wrong result.

PR #1310, opened 2024-10-25
Resolve scgi software crash

stickz:
This commit resolves a scgi software crash when the scgi socket is closed before the message can be sent. It instructs `::send()` not to send a `SIGPIPE` termination signal. Instead the value `-1` is returned and handled bellow. The `SCgiTask` is closed.

```
Thread 3 "rtorrent scgi" received signal SIGPIPE, Broken pipe.
                                                             [Switching to Thread 0x7fffe635c6c0 (LWP 2443872)]
0x00007ffff7929a84 in send () from /lib/x86_64-linux-gnu/libc.so.6
```

### DIFF
--- a/src/rpc/scgi_task.cc
+++ b/src/rpc/scgi_task.cc
@@ -200,7 +200,7 @@ SCgiTask::event_read() {
 
 void
 SCgiTask::event_write() {
-  int bytes = ::send(m_fileDesc, m_position, m_bufferSize, 0);
+  int bytes = ::send(m_fileDesc, m_position, m_bufferSize, MSG_NOSIGNAL);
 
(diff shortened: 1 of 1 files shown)


PPN-SD:
Well, this patch causes a failure on my scgi socket, unreachable (rutorrent).

stickz:
> Well, this patch causes a failure on my scgi socket, unreachable (rutorrent).

If rtorrent is crashing please provide a stack trace. The unreachable message on ruTorrent is a crash.

This commit stops rtorrent from raising an exception which triggers a crash. Saying it causes a failure is equivalent to saying it does nothing. This is a truth.

PR #1312, opened 2024-11-10
Add clang-format/clang-tidy config files

kannibalox:
I chose the rules to try and be closest aligned to the existing format, and the three additional files have been `clang-format -i`'d to be representative of the changes it would make across the codebase.

I'm unsure of your opinion on code formatters, so here's my personal quick take:
Advantages:
 * New contributors don't have to figure out the format based on existing code. For me personally, I put in a real effort to try and emulate it, and I still got a lot of things wrong.
 * You'd no longer have the cognitive load of having to correct people using a different style

Disadvantages:
 * Large commits when initially changing the style. This can be somewhat mitigated by only formatting on new changes, instead of the whole codebase at once.
 * It'll never match your personal style perfectly. As a counterpoint, to steal a quote from [Rob Pike](https://x.com/bitfield/status/953395343353315329): clang-format's style is nobody's favorite, but clang-format is everybody's favorite

Even if you decide against it, I've found these rules helpful for catching variable naming convention misses, so hopefully it might serve as a reference for other contributors.

### DIFF
--- a/src/command_dynamic.cc
+++ b/src/command_dynamic.cc
@@ -5,12 +5,12 @@
 // it under the terms of the GNU General Public License as published by
 // the Free Software Foundation; either version 2 of the License, or
 // (at your option) any later version.
-// 
+//
 // This program is distributed in the hope that it will be useful,
 // but WITHOUT ANY WARRANTY; without even the implied warranty of
 // MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 // GNU General Public License for more details.
-// 
+//
 // You should have received a copy of the GNU General Public License
 // along with this program; if not, write to the Free Software
 // Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
@@ -40,25 +40,23 @@
 #include <torrent/utils/log.h>
 #include <torrent/utils/option_strings.h>
 
-#include "globals.h"
-#include "control.h"
 #include "command_helpers.h"
+#include "control.h"
+#include "globals.h"
 #include "rpc/parse.h"
 #include "rpc/parse_options.h"
 
-static std::vector<std::pair<const char*, int>> object_storage_flags = {
-  { "multi", rpc::object_storage::flag_multi_type },
-  { "simple", rpc::object_storage::flag_function_type },
-  { "value", rpc::object_storage::flag_value_type },
-  { "bool", rpc::object_storage::flag_bool_type },
-  { "string", rpc::object_storage::flag_string_type },
-  { "list", rpc::object_storage::flag_list_type },
+static std::vector<std::pair<const char*, int>> object_storage_flags = {{"multi", rpc::object_storage::flag_multi_type},
+                                                                        {"simple", rpc::object_storage::flag_function_type},
+                                                                        {"value", rpc::object_storage::flag_value_type},
+                                                                        {"bool", rpc::object_storage::flag_bool_type},
+                                                                        {"string", rpc::object_storage::flag_string_type},
+                                                                        {"list", rpc::object_storage::flag_list_type},
 
-  { "static", rpc::object_storage::flag_static },
-  { "private", rpc::object_storage::flag_private },
-  { "const", rpc::object_storage::flag_constant },
-  { "rlookup", rpc::object_storage::flag_rlookup }
-};
+                                                                        {"static", rpc::object_storage::flag_static},
+                                                                        {"private", rpc::object_storage::flag_private},
+                                                                        {"const", rpc::object_storage::flag_constant},
+                                                                        {"rlookup", rpc::object_storage::flag_rlookup}};
 
 static int
 object_storage_parse_flag(const std::string& flag) {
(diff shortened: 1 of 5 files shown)


src/command_dynamic.cc
```diff
@@ -40,25 +40,23 @@
...
-  { "value", rpc::object_storage::flag_value_type },
-  { "bool", rpc::object_storage::flag_bool_type },
-  { "string", rpc::object_storage::flag_string_type },
-  { "list", rpc::object_storage::flag_list_type },
+static std::vector<std::pair<const char*, int>> object_storage_flags = {{"multi", rpc::object_storage::flag_multi_type},
+                                                                        {"simple", rpc::object_storage::flag_function_type},
+                                                                        {"value", rpc::object_storage::flag_value_type},
+                                                                        {"bool", rpc::object_storage::flag_bool_type},
+                                                                        {"string", rpc::object_storage::flag_string_type},
+                                                                        {"list", rpc::object_storage::flag_list_type},
```
rakshasa, line 54: Is there really no option to prefer the former?

src/command_dynamic.cc
```diff
@@ -299,14 +312,11 @@ system_method_insert(const torrent::Object::list_type& args) {
...
-      (new_flags & rpc::object_storage::flag_multi_type)) {
+  if ((new_flags & rpc::object_storage::flag_function_type) || (new_flags & rpc::object_storage::flag_multi_type)) {
     new_args.push_back(system_method_generate_command(++itrArgs, args.end()));
 
-  } else if ((new_flags & rpc::object_storage::flag_value_type) ||
-             (new_flags & rpc::object_storage::flag_bool_type) ||
-             (new_flags & rpc::object_storage::flag_string_type) ||
-             (new_flags & rpc::object_storage::flag_list_type)) {
+  } else if ((new_flags & rpc::object_storage::flag_value_type) || (new_flags & rpc::object_storage::flag_bool_type) ||
+             (new_flags & rpc::object_storage::flag_string_type) || (new_flags & rpc::object_storage::flag_list_type)) {
```
rakshasa, line 319: This is (very) bad style imo.

src/command_dynamic.cc
```diff
@@ -345,8 +355,10 @@ system_method_redirect(const torrent::Object::list_type& args) {
   std::string new_key  = torrent::object_create_string(args.front());
   std::string dest_key = torrent::object_create_string(args.back());
 
-  rpc::commands.create_redirect(create_new_key(new_key), create_new_key(dest_key),
-                                rpc::CommandMap::flag_public_xmlrpc | rpc::CommandMap::flag_delete_key | rpc::CommandMap::flag_modifiable);
+  rpc::commands.create_redirect(create_new_key(new_key),
+                                create_new_key(dest_key),
+                                rpc::CommandMap::flag_public_xmlrpc | rpc::CommandMap::flag_delete_key |
+                                  rpc::CommandMap::flag_modifiable);
```
rakshasa, line 361: 70 char lines is a relic of ancient coding practices, it doesn't make sense for such verbose and namespace heavy cases as this.

Breaking up lines like this just makes it worse.

Should either use newline for all, keep it on one line or respect the choice made by the programmer.

src/core/view.cc
```diff
@@ -91,9 +91,7 @@ struct view_downloads_compare : std::binary_function<Download*, Download*, bool>
 struct view_downloads_filter : std::unary_function<Download*, bool> {
   view_downloads_filter(const torrent::Object& cmd, const torrent::Object& cmd2) : m_command(cmd), m_command2(cmd2) {}
 
-  bool operator () (Download* d1) const {
-    return this->evalCmd(m_command, d1) && this->evalCmd(m_command2, d1);
-  }
+  bool operator()(Download* d1) const { return this->evalCmd(m_command, d1) && this->evalCmd(m_command2, d1); }
```
rakshasa, line 94: This is less readable, and should respect the choice of the programmer.

kannibalox:
I was using `ColumnLimit: 130` which is responsible for a lot of the newlines shenanigans, which I arrived at by starting at 80 and bumping up by 10 until it was least weird. I did try `ColumnLimit: 0`, which as intended respected existing newlines, but that caused weirdness in other areas, e.g.:

```diff
-      rpc::commands.insert_slot<rpc::command_base_is_type<rpc::command_base_call_value<rpc::target_type>>::type>                                                                   
-       (create_new_key<5>(rawKey, ".set"),                                                                                                                                          
-        std::bind(&rpc::object_storage::set_str_bool, control->object_storage(), rawKey, std::placeholders::_2),                                                                    
-        &rpc::command_base_call_value<rpc::target_type>,                                                                                                                            
-        cmd_flags,                                                                                                                                                                  
-        NULL,                                                                                                                                                                       
-        NULL);                                                                                                                                                                      
+      rpc::commands.insert_slot<rpc::command_base_is_type<rpc::command_base_call_value<rpc::target_type>>::type>(create_new_key<5>(rawKey, ".set"),                                 
+                                                                                                                 std::bind(&rpc::object_storage::set_str_bool, control->object_stor$
+                                                                                                                 &rpc::command_base_call_value<rpc::target_type>,                   
+                                                                                                                 cmd_flags,                                                         
+                                                                                                                 NULL,                                                              
+                                                                                                                 NULL);
```
I think I was just trying to hard to be a purist though, because those can be fixed by just manually moving the `(`, and `ColumnLimit: 0` will respect that.

tl;dr @rakshasa it should respect existing newlines now, unless other rules override them

Issue #1313, opened 2024-11-12
significant performance / usability regression from fdatasync(2) change


slingamn:
d4427931a2f03b8fd9f80895ace12fb95e0b759c added an fdatasync(2) for each metadata update. Each such operation can be quite slow on HDDs (I'm measuring them at 15-30 milliseconds per operation with `strace -T`). With two such operations per active torrent, and 1000 active torrents, the main thread (including the UI and all upload and download activity) will freeze for approximately a minute during the periodic `session_save` operation:

https://github.com/rakshasa/rtorrent/blob/ee120730cc046a9628bcd5d26f8a1f9b99fab158/src/main.cc#L358

resulting in only 95% availability for the client, since the operation is scheduled every 20 minutes.

I'm wondering if there could be a better way to handle this? Some possible ideas that come to mind:

1. Trigger sync(2) on a separate thread after all metadata updates are complete
2. fdatasync(2) on individual files, but on a separate thread via some kind of message passing
3. Make the file format more resilient to corruption (although that might not resolve all the concerns mentioned in https://github.com/jesec/rtorrent/commit/f9a875b633eb083ec82f89ebeae8678ca424e8b2 )

Thanks for your time.

rakshasa:
The issue is that we need to wait for the files to be sync'ed before calling rename.

This makes it easy to fix the blocking issue as the rename can be done lazily in a separate thread.

uis246:
3. You still would need to send barrier to OS. Otherwise writes can be reordered in whatever way OS and hardware feel.

uis246:
I'm only learning about rtorrent. But here are my thoughts:
1. If there are multiple files to be synced, then writing them all and then batch-sync would be faster.
2. If there is only one file to be synced, then move `fdatasync()` before closing file. Not that opening and closing are too expensive on Linux.

rakshasa:
I'll look into making this a priority after finishing the tracker changes.

PR #1316, opened 2024-11-13
Keep a cache of session file data, and only write if changes occur

kannibalox:
This is mainly in response to https://github.com/rakshasa/rtorrent/issues/1313, but making `session.save` less heavy in general has been something I've thought about, and this approach seemed simpler than trying to mark downloads as dirty or have logic depending on the state as described [here](https://github.com/rakshasa/libtorrent/blob/master/TODO_LONGTERM#L41-L43). The overhead of a cache miss is pretty low, but https://github.com/rakshasa/libtorrent/pull/257 was mainly to make this much more effective for libtorrent_resume files. I'm on the fence about whether this is an elegant use of cache or a terrible hack, so I submit it for your consideration.

It includes two other changes to behavior:
* It's now possible for `.libtorrent_resume` to get updated while `.rtorrent` doesn't. I couldn't think of a scenario where this would be bad, and it would've complicated error handling, but it can be done if desired.
* An `::access()` call is made on a cache hit. This is primarily to support the rare but important use case of being able to restore a session directory entirely from scratch on a running instance (e.g. if someone accidentally deletes it)

Something to consider might be adding an override flag so that `d.save_full_session` and `d.save_resume` can ignore the cache.

### DIFF
--- a/src/core/download_store.cc
+++ b/src/core/download_store.cc
@@ -38,7 +38,9 @@
 
 #include "config.h"
 
+#include <string>
 #include <fstream>
+#include <sstream>
 #include <stdio.h>
 #include <fcntl.h>
 #include <unistd.h>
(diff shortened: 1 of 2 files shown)


rakshasa:
There's no reason why we should need to do work in the main thread, we only need to pass encoded bencode blobs to a worker thread and have it handle everything.

As such I don't think cache's should be added before that is done.

kannibalox:
Makes sense, I'll close this for now and can revisit it when things are threaded.

PR #1317, opened 2024-11-15
Fix calls ar directly

PPN-SD:
See : https://bugs.gentoo.org/943039 and https://github.com/gentoo/gentoo/pull/39256

Toolchain tools called directly cause issue in some environnements (cross-compiling, among others)


stickz:
@rakshasa This is an important PR for Gentoo Linux. They are waiting for upstream review to commit this stability patch.

PR #1324, opened 2024-12-01
Convert __sync* built-ins to std::atomic

kannibalox:
Also converts thread_queue_hack to a simpler reserved vector with a std::mutex.

Sibling PR to https://github.com/rakshasa/libtorrent/pull/265, addresses https://github.com/rakshasa/libtorrent/issues/253

### DIFF
--- a/src/control.h
+++ b/src/control.h
@@ -37,6 +37,7 @@
 #ifndef RTORRENT_CONTROL_H
 #define RTORRENT_CONTROL_H
 
+#include <atomic>
 #include <cinttypes>
 #include <sys/types.h>
 #include <rak/timer.h>
@@ -87,8 +88,8 @@ public:
 
   void                handle_shutdown();
 
-  void                receive_normal_shutdown()     { m_shutdownReceived = true; __sync_synchronize(); }
-  void                receive_quick_shutdown()      { m_shutdownReceived = true; m_shutdownQuick = true; __sync_synchronize(); }
+  void                receive_normal_shutdown()     { m_shutdownReceived = true; }
+  void                receive_quick_shutdown()      { m_shutdownReceived = true; m_shutdownQuick = true; }
 
   core::Manager*      core()                        { return m_core; }
   core::ViewManager*  view_manager()                { return m_viewManager; }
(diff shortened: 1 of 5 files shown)


src/control.h
```diff
@@ -134,8 +135,8 @@ class Control {
 
   rak::priority_item  m_taskShutdown;
 
-  bool                m_shutdownReceived lt_cacheline_aligned;
-  bool                m_shutdownQuick lt_cacheline_aligned;
+  std::atomic<bool>   m_shutdownReceived lt_cacheline_aligned;
+  std::atomic<bool>   m_shutdownQuick lt_cacheline_aligned;
```
rakshasa, line 139: These are cacheline aligned, so we should remove all instances of `lt_cacheline_aligned` when we're not doing `__sync`.

src/thread_base.cc
```diff
@@ -1,122 +1,58 @@
...
+    std::lock_guard<std::mutex> lguard(m_lock);
+    target.assign(begin(), end());
+    clear();
   }
 
- private:
-  int        m_lock;
-  value_type m_queue[max_size + 1];
+private:
+  std::mutex m_lock;
```
rakshasa, line 41: `empty()` isn't lock protected.

src/thread_base.cc
```diff
@@ -1,122 +1,61 @@
...
-
-    while (*itr != NULL) *dest++ = *itr++;
-
-    clear_and_unlock();
-    return dest;
+  void push_back(value_type v) {
+    if (size() >= max_size)
+      throw torrent::internal_error("Overflowed thread_queue max size of " + std::to_string(max_size) + ".");
+    const std::lock_guard<std::mutex> lguard(m_lock);
+    base_type::push_back(v);
```
rakshasa, line 34: Lock before size call, also use `base_type::size()` to be explicit.

src/thread_base.cc
```diff
@@ -1,122 +1,61 @@
...
 #include "globals.h"
-#include "control.h"
-#include "core/manager.h"
 
-// Temporarly injected into config.h.
-/* temp hack */
-//#define lt_cacheline_aligned __attribute__((__aligned__(128)))
-
-class lt_cacheline_aligned thread_queue_hack {
+class lt_cacheline_aligned thread_queue_hack : private std::vector<ThreadBase::thread_base_func> {
```
rakshasa, line 16: Remove cacheline.

src/thread_worker.cc
```diff
@@ -67,17 +67,13 @@ ThreadWorker::init_thread() {
 
 bool
 ThreadWorker::set_scgi(rpc::SCgi* scgi) {
-  if (!__sync_bool_compare_and_swap(&m_safe.scgi, NULL, scgi))
+  if (m_scgi != nullptr)
     return false;
 
-  change_xmlrpc_log();
+  m_scgi = scgi;
```
rakshasa, line 73: This should use `compare_exchange_strong`.

PR #1326, opened 2024-12-04
fix: `d.group.name` should return group name instead of group index

trim21:
I'm new to C++, sorry if it's not correct.

### DIFF
--- a/src/command_download.cc
+++ b/src/command_download.cc
@@ -905,7 +905,7 @@ initialize_command_download() {
   //                                           CG_GROUP_INDEX()));
 
   CMD2_DL         ("d.group",      std::bind(&cg_d_group, std::placeholders::_1));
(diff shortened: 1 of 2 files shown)


rakshasa:
Post logs of when you're calling the command with and without this change.

trim21:
I didn't use tui but header less server with xmlrpc what log do you want? The output of this command?

rakshasa:
Yes, just the output of the xmlrpc commands.

trim21:
first create a download and get it's info_hash, then:

```py
client.xml_call("choke_group.insert", ["", "test"])
client.xml_call("d.group.set", [info_hash.upper(), "test"])
group_name = client.xml_call("d.group.name", [info_hash.upper()])
print(type(group_name), repr(group_name))
```

before it's `<class 'int'> 1` and after it's `<class 'str'> 'test'`

trim21:
Do we have a plan for a new release?

PR #1328, opened 2024-12-05
scgi_task: allow compiling when MSG_NOSIGNAL undefined

barracuda156:


barracuda156:
Everything else works now without patching, this is the only issue:
```
rpc/scgi_task.cc: In member function 'virtual void rpc::SCgiTask::event_write()':
rpc/scgi_task.cc:203:60: error: 'MSG_NOSIGNAL' was not declared in this scope
  203 |   int bytes = ::send(m_fileDesc, m_position, m_bufferSize, MSG_NOSIGNAL);
      |                                                            ^~~~~~~~~~~~
make[2]: *** [rpc/scgi_task.o] Error 1
```

rakshasa:
This is an issue with your libcurl version or something, and that flag is vital.

barracuda156:
No, it is absent in the system. See, for example: https://github.com/lpeterse/haskell-socket/issues/8
Solaris also does not have it.

stickz:
@barracuda156 @rakshasa I have submitted #1329 with a proper fix. It will detect Apple or Solaris at compile time and only disable the flag for these platforms. It's undefined anyways, so it's not going to accomplish anything.

PR #1329, opened 2024-12-05
scgi: Fix Apple and Solaris compatibility

stickz:
This is a follow up to #1310. Apple and Solaris do not support `MSG_NOSIGNAL`, so disable this flag for these platforms until we find a better solution. Other platforms remain unaffected due to the compile time check.


rakshasa:
I'd rather have it be a proper autoconf check as this kind of check buried deep in the code will be forgotten, and if those archs or libcurl start supporting it will be an issue.

Issue #1302, opened 2024-10-02
Segfault while exiting rtorrent with an active download


xsmile:
# Description

rtorrent produces a segfault if the application is exited while a torrent is being downloaded.

# Additional info
 
* OS: Arch Linux
* tested rtorrent versions: 0.10.0, 0.9.8, 0.9.6
* curl version: 8.10.1

# Steps to reproduce

1. add and start downloading a torrent
2. exit rtorrent

# Details

```
(gdb) bt
#0  core::CurlSocket::close (this=0x5555559f1aa0) at core/curl_socket.cc:103
#1  core::CurlSocket::receive_socket (easy_handle=<optimized out>, fd=<optimized out>, what=<optimized out>, userp=<optimized out>, socketp=0x5555559f1aa0) at core/curl_socket.cc:62
#2  0x00007ffff7ea58f6 in Curl_multi_closed (data=0x5555556b0be0, s=14) at /usr/src/debug/curl/curl/lib/multi.c:3035
#3  0x00007ffff7ef05bd in socket_close.isra.0 (data=0x5555556b0be0, conn=0x555555a699b0, use_callback=<optimized out>, sock=14) at /usr/src/debug/curl/curl/lib/cf-socket.c:424
#4  0x00007ffff7e5fc47 in cf_socket_close (cf=0x555555a7b410, data=<optimized out>) at /usr/src/debug/curl/curl/lib/cf-socket.c:1005
#5  0x00007ffff7e69e56 in cf_he_close (cf=0x555555b92b30, data=0x5555556b0be0) at /usr/src/debug/curl/curl/lib/connect.c:1004
#6  0x00007ffff7eddef4 in ssl_cf_close (cf=0x555555b73760, data=0x5555556b0be0) at /usr/src/debug/curl/curl/lib/vtls/vtls.c:1666
#7  0x00007ffff7e69f0e in cf_setup_close (cf=0x555555c533a0, data=0x5555556b0be0) at /usr/src/debug/curl/curl/lib/connect.c:1359
#8  0x00007ffff7e5e6e2 in cf_hc_close (cf=0x555555a8e5d0, data=0x5555556b0be0) at /usr/src/debug/curl/curl/lib/cf-https-connect.c:496
#9  0x00007ffff7e69f7e in Curl_conn_close (data=data@entry=0x5555556b0be0, index=index@entry=0) at /usr/src/debug/curl/curl/lib/cfilters.c:180
[...]

PR #1335, opened 2024-12-10
Ensure thread is still active when closing a curl socket

kannibalox:
This seems like a reasonable check since there's no guarantee that the torrent::manager will be non-`NULL` as long as this ordering is preserved:
https://github.com/rakshasa/rtorrent/blob/7e1193acabc4a10417aabdeced60eda9e5621733/src/core/manager.cc#L197-L199

On the other hand, I can only reproduce the issue with IPv6 trackers (on a non-IPv6 enabled network), and there's the possibility of finding a more general IPv6 fix elsewhere.

Fixes https://github.com/rakshasa/rtorrent/issues/1302


rakshasa:
#1338

PR #1338, opened 2024-12-12
Fixed curl stack shutdown with active downloads.

rakshasa:


kannibalox:
Can confirm, this fixes the issue as well.

PR #1339, opened 2024-12-13
Add page up/down and home/end bindings to download list

### LINKED ISSUE #222: [Feature request] Support for Page Up/Page Down
Hi,

Browsing long lists is a bit tedious. Would it be possible to implement support for page up/page down? Or at least an equivalent shortcut with the ctrl-modifier?

kannibalox:
The choice of `50` as page size is pretty much arbitrary. I assumed `// These should also be commands.` only applied to the view change bindings, but I can make these commands as well if desired.

Closes #222

### DIFF
--- a/src/core/view.cc
+++ b/src/core/view.cc
@@ -1,39 +1,3 @@
-// rTorrent - BitTorrent client
-// Copyright (C) 2005-2011, Jari Sundell
-//
-// This program is free software; you can redistribute it and/or modify
-// it under the terms of the GNU General Public License as published by
-// the Free Software Foundation; either version 2 of the License, or
-// (at your option) any later version.
-//
-// This program is distributed in the hope that it will be useful,
-// but WITHOUT ANY WARRANTY; without even the implied warranty of
-// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
-// GNU General Public License for more details.
-//
-// You should have received a copy of the GNU General Public License
-// along with this program; if not, write to the Free Software
-// Foundation, Inc., 59 Temple Place, Suite 330, Boston, MA  02111-1307  USA
-//
-// In addition, as a special exception, the copyright holders give
-// permission to link the code of portions of this program with the
-// OpenSSL library under certain conditions as described in each
-// individual source file, and distribute linked combinations
-// including the two.
-//
-// You must obey the GNU General Public License in all respects for
-// all of the code used other than OpenSSL.  If you modify file(s)
-// with this exception, you may extend this exception to your version
-// of the file(s), but you are not obligated to do so.  If you do not
-// wish to do so, delete this exception statement from your version.
-// If you delete this exception statement from all source files in the
-// program, then also delete it here.
-//
-// Contact:  Jari Sundell <jaris@ifi.uio.no>
-//
-//           Skomakerveien 33
-//           3185 Skoppum, NORWAY
-
 #include "config.h"
 
 #include <algorithm>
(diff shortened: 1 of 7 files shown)


src/ui/element_download_list.cc
```diff
@@ -67,40 +31,42 @@ ElementDownloadList::ElementDownloadList() :
...
+
+  m_bindings[KEY_UP] = m_bindings['P' - '@'] = std::bind(&ElementDownloadList::receive_prev, this);
   m_bindings[KEY_DOWN] = m_bindings['N' - '@'] = std::bind(&ElementDownloadList::receive_next, this);
 
-  m_bindings['L']           = std::bind(&ElementDownloadList::toggle_layout, this);
+  m_bindings[KEY_PPAGE] = [this] { receive_prev_page(); };
+  m_bindings[KEY_NPAGE] = [this] { receive_next_page(); };
+
+  m_bindings[KEY_HOME] = [this] { receive_home(); };
+  m_bindings[KEY_END]  = [this] { receive_end(); };
```
rakshasa, line 67: Include bindings for these; ^P, ^N, and something for home/end.

src/core/view.cc
```diff
@@ -237,20 +201,20 @@ View::set_not_visible(Download* download) {
...
 }
 
 void
-View::prev_focus() {
+View::prev_focus(int i) {
   if (empty())
     return;
 
-  m_focus = (m_focus - 1 + size() + 1) % (size() + 1);
+  m_focus = (m_focus - i + size() + 1) % (size() + 1);
```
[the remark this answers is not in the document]
kannibalox: Done.

src/ui/element_download_list.cc
kannibalox: ^P and ^N are taken for up/down, and I couldn't find a good consensus, so went with ^J and ^K.

I used ^A and ^E for home/end, that's pretty standard (e.g. readline/emacs)

src/core/view.cc
```diff
@@ -237,20 +201,41 @@ View::set_not_visible(Download* download) {
...
+    m_focus = size();
+  } else {
+    // Move forward, stop at the boundary
+    if (m_focus == size())
+      m_focus = i - 1;
+    else
+      m_focus += i;
+    if (m_focus > size() - 1)
+      m_focus = size() - 1;
+  }
```
rakshasa, line 219: This looks wrong and is messy.

```
if (empty() || i == 0)
  return;

if (m_focus == size() - 1) {
  m_focus = 0;
  emit_changed();
  return;
}

m_focus += i;

if (m_focus > size() - 1)
  m_focus = size() - 1;

emit_changed();
```

Same for prev_focus.

kannibalox: Maybe I misunderstood, but this is the logic required to support the following UX for a page down of 50 with 146 items:
`none (default state) -> 50 -> 100 -> 146 ("next" boundary) -> none`
For page up:
`none -> 146 -> 96 -> 46 -> none ("prev" boundary)`
while keeping KEY_DOWN/KEY_UP the same. I did try that code early on and found that it will never be able to reach m_focus == size(), which makes KEY_DOWN work differently than it currently does. If you want to change the flow so that page up/down are the ones who can never reach size(), I can split things out into two functions.

I have fixed up the style to match, though, and moved ^K to ^U to fix a key binding conflict.

rakshasa: Should probably have both be able to reach `size()`, kind of forgot how it worked.

kannibalox:
Updated to use a new value command, `ui.focus.page_size`, which if not set to >0, rtorrent will auto-detect the page size based on how big the current display is

PR #1341, opened 2024-12-16
Create single-element maps for functions instead of using dict_keys

kannibalox:
Requires


rakshasa:
I don't like the use of a single-item map here, it is too easy to use wrongly and the dict type is a special type for method+args and ordered key+value lists.

Instead this should be a new class in libtorrent that holds a pointer to a inherited std::vector with added key string.

PR #1348, opened 2024-12-24
tinyxml2: Correctly handle commands that are flagged as not using targets

### LINKED ISSUE #1346: ruTorrent: Faults with tinyxml2
@kannibalox Do you know why tinyxml2 is faulting on ruTorrent? Here is my rpc log.

```
[2024-12-24 15:25:41] <?xml version="1.0" encoding="UTF-8"?><methodCall><methodName>to_kb</methodName><params>
<param><value><i8>1024</i8></value></param>
</params></methodCall>
[2024-12-24 15:25:41] Status: 200 OK
Content-Type: text/xml
Content-Length: 269

<?xml version="1.0"?><methodResponse><fault><struct><member><name>faultCode</name><value><i4>-500</i4></value></member><member><name>faultString</name><value><string>invalid parameters: target must be a string</string></value></member></struct></fault></methodResponse>
```

kannibalox:
Fixes #1346

### DIFF
--- a/src/rpc/xmlrpc.cc
+++ b/src/rpc/xmlrpc.cc
@@ -59,6 +59,9 @@ private:
 
 void
 XmlRpc::object_to_target(const torrent::Object& obj, int callFlags, rpc::target_type* target) {
+  if (callFlags & CommandMap::flag_no_target)
+    return;
+
   if (!obj.is_string()) {
     throw torrent::input_error("invalid parameters: target must be a string");
   }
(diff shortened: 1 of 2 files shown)


stickz:
Works great thanks @kannibalox! I'm working on deploying your tinyxml2 changes to thousands of clients on crazy max docker. That's what the GitHub spam is about! I'll let you know if we find anymore issues.

rakshasa:
Really should start deprecating old no-target commands.

PR #1355, opened 2024-12-28
tinyxml2: in xmlrpc an array must include values in a data element

simonc56:
`tinyxml2` implementation of xmlrpc from https://github.com/rakshasa/rtorrent/pull/1322 does not follow XML-RPC [spec](https://xmlrpc.com/spec.md) concerning array :
> An < array > contains a single < data > element, which can contain any number of < value >s.

This PR adds the `<data>` tag in `<array>`.

Fixes :
- https://github.com/crazy-max/docker-rtorrent-rutorrent/issues/419


kannibalox:
Obviously this is on me for not following the spec, but I'm amazed that it's been tested with at least 3 distinct XML-RPC clients without this getting flagged.

PR #1373, opened 2025-01-12
Add JSON-RPC capability

kannibalox:
Based off the work in https://github.com/jesec/rtorrent, this implements the 2.0 specification: https://www.jsonrpc.org/specification. It adds a new RpcManager class and a little SCGI logic to handle switching between XML and JSON seamlessly. If the `CONTENT_TYPE` header is present and set to `application/json` (or `text/xml` for XML-RPC), it'll use that, otherwise it will peek at the body to auto-detect it.

The main benefit is reduced message size, as seen on this synthetic set of 10k torrents:
```
$ rtxmlrpc -U scgi://localhost:8888?rpc=xml d.multicall2 '' main d.name= d.hash= d.size_bytes= --debug |& grep stats
DEBUG:pyrosimple.scripts.rtxmlrpc.RtorrentXmlRpc:RPC stats: 1 requests (387 bytes) in 0.029s (response 1.9 MiB)

$ rtxmlrpc -U scgi://localhost:8888?rpc=json d.multicall2 '' main d.name= d.hash= d.size_bytes= --debug |& grep stats
DEBUG:pyrosimple.scripts.rtxmlrpc.RtorrentXmlRpc:RPC stats: 1 requests (106 bytes) in 0.024s (response 710.9 KiB)
```

One limitation of JSON-RPC is that it doesn't have a dedicated binary data type. Due to that, [base64 data URI](https://developer.mozilla.org/en-US/docs/Web/URI/Schemes/data) support has been added to all the `load.*` commands. Those are currently the only commands which normally need to take binary data, but any future commands would have to follow the same pattern or implement their own scheme.

<details>
  <summary>Note for Flood users</summary>
  When Flood detects the presence of JSON-RPC, it assumes that the full jesec/rtorrent capabilities are available, which breaks a couple things. Thankfully, it's pretty easy to provide some stubs/redirects to get it working:

```ini
method.redirect=load.throw,load.normal
method.redirect=load.start_throw,load.start
method.insert=d.down.sequential,value|const,0
method.insert=d.down.sequential.set,value|const,0
```
</details>

### DIFF
--- a/src/command_network.cc
+++ b/src/command_network.cc
@@ -95,7 +95,7 @@ apply_tos(const torrent::Object::string_type& arg) {
 torrent::Object apply_encoding_list(const std::string& arg) { torrent::encoding_list()->push_back(arg); return torrent::Object(); }
 
 torrent::File*
-xmlrpc_find_file(core::Download* download, uint32_t index) {
+rpc_find_file(core::Download* download, uint32_t index) {
   if (index >= download->file_list()->size_files())
     return NULL;
 
@@ -104,7 +104,7 @@ xmlrpc_find_file(core::Download* download, uint32_t index) {
 
 // Ergh... time to update the Tracker API to allow proper ptrs.
 torrent::Tracker*
-xmlrpc_find_tracker(core::Download* download, uint32_t index) {
+rpc_find_tracker(core::Download* download, uint32_t index) {
   if (index >= download->tracker_list()->size())
     return NULL;
 
@@ -112,7 +112,7 @@ xmlrpc_find_tracker(core::Download* download, uint32_t index) {
 }
 
 torrent::Peer*
-xmlrpc_find_peer(core::Download* download, const torrent::HashString& hash) {
+rpc_find_peer(core::Download* download, const torrent::HashString& hash) {
   torrent::ConnectionList::iterator itr = download->connection_list()->find(hash.c_str());
 
   if (itr == download->connection_list()->end())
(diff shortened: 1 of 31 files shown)


src/command_network.cc
```diff
@@ -122,32 +122,31 @@ xmlrpc_find_peer(core::Download* download, const torrent::HashString& hash) {
...
-  rpc::xmlrpc.slot_find_download() = std::bind(&core::DownloadList::find_hex_ptr, control->core()->download_list(), std::placeholders::_1);
-  rpc::xmlrpc.slot_find_file() = std::bind(&xmlrpc_find_file, std::placeholders::_1, std::placeholders::_2);
-  rpc::xmlrpc.slot_find_tracker() = std::bind(&xmlrpc_find_tracker, std::placeholders::_1, std::placeholders::_2);
-  rpc::xmlrpc.slot_find_peer() = std::bind(&xmlrpc_find_peer, std::placeholders::_1, std::placeholders::_2);
+initialize_rpc() {
+  rpc::rpc.initialize(
+                      [](const char* hash) { return control->core()->download_list()->find_hex_ptr(hash); },
+                      [](core::Download* d, uint32_t index) { return rpc_find_file(d, index); },
+                      [](core::Download* d, uint32_t index) { return rpc_find_tracker(d, index); },
+                      [](core::Download* d, const torrent::HashString& hash) { return rpc_find_peer(d, hash); });
```
rakshasa, line 130: Not a fan of this style without assignment being done to named slots.

src/core/manager.cc
```diff
@@ -333,10 +340,20 @@ Manager::try_create_download(const std::string& uri, int flags, const command_li
...
   f->slot_finished([f]() { delete f; });
 
-  if (flags & create_raw_data)
+  if (is_data_uri(uri)) {
+    // Allow the use of data URIs, primarily for JSON-RPC which
+    // doesn't have a defined mechanism for binary data
+    const unsigned long start = uri.find("base64,", 5) + 7;
+    if (start >= uri.size())
+      throw torrent::input_error("Empty base64.");
+    auto output = utils::decode_base64(uri.substr(start));
```
rakshasa, line 349: Create a function `decode_data_uri` that does both the check and decoding.

src/rpc/rpc_manager.cc
```diff
@@ -0,0 +1,140 @@
+#include "config.h"
+
+#include <cstring>
+
+#include <torrent/exceptions.h>
+
+#include "parse_commands.h"
+
+#include "rpc/rpc_manager.h"
```
rakshasa, line 9: Unneeded empty newlines.

rakshasa:
This should probably add options to enable/disable xmlrpc or json handling.

kannibalox:
> This should probably add options to enable/disable xmlrpc or json handling.

Were you thinking compile flag options or command options, and can you expand on the use case for this? They should be able to coexist peacefully, if a client doesn't want to use one or the other all they have to do is not send that type of request.

rakshasa:
One example is the flood users, who knows where else this might cause issues.

This allows users/providers force the use of a particular protocol for performance or compatibility reasons.

It should be command value to enable/disable each.

kannibalox:
Added a toggle for each, and renamed some of the commands/variables from `xmlrpc` to `rpc`

src/command_network.cc
```diff
@@ -298,9 +297,12 @@ initialize_command_network() {
...
 
-  CMD2_ANY_STRING  ("network.xmlrpc.dialect.set",    std::bind(&apply_xmlrpc_dialect, std::placeholders::_2));
-  CMD2_ANY         ("network.xmlrpc.size_limit",     std::bind(&rpc::XmlRpc::size_limit, rpc::xmlrpc));
-  CMD2_ANY_VALUE_V ("network.xmlrpc.size_limit.set", std::bind(&rpc::XmlRpc::set_size_limit, rpc::xmlrpc, std::placeholders::_2));
+  CMD2_ANY_STRING  ("network.xmlrpc.dialect.set",    [](const auto&, const auto& arg) { return apply_xmlrpc_dialect(arg); })
+  CMD2_ANY         ("network.xmlrpc.size_limit",     [](const auto&, const auto&){ return rpc::rpc.size_limit(); });
+  CMD2_ANY_VALUE_V ("network.xmlrpc.size_limit.set", [](const auto&, const auto& arg){ return rpc::rpc.set_size_limit(arg); });
+
+  CMD2_VAR_BOOL    ("network.rpc.xmlrpc_enabled",    true);
+  CMD2_VAR_BOOL    ("network.rpc.jsonrpc_enabled",   true);
```
rakshasa, line 305: `network.rpc.use_xmlrpc`, etc

PPN-SD:
@kannibalox, I get two QA warnings with it : 

>  * rpc/rpc_manager.cc:105:1: warning: control reaches end of non-void function [-Wreturn-type]
>  * rpc/rpc_manager.cc:129:1: warning: control reaches end of non-void function [-Wreturn-type]

Abzie:
> It should be command value to enable/disable each.

> Added a toggle for each, and renamed some of the commands/variables from `xmlrpc` to `rpc`

Can the documentation be updated to show how to configure rtorrent.rc to use either XMLRPC or JSON-RPC?

What is the default behaviour? JSON-RPC? XMLRPC? Both?

This change, present in 0.15.2, is breaking rutorrent (personally tested on versions 5.1.6 and 5.1.7).

PR #1376, opened 2025-01-15
cleanup: Changes for c++17

stickz:
This cleanup adjusts the rTorrent code to comply with c++17 standard. It also simplifies various parts of the code.

### DIFF
--- a/src/input/manager.cc
+++ b/src/input/manager.cc
@@ -64,7 +64,7 @@ Manager::pressed(int key) {
   if (m_textInput != NULL)
     m_textInput->pressed(key);
   else
-    std::find_if(rbegin(), rend(), std::bind2nd(std::mem_fun(&Bindings::pressed), key));
+    std::find_if(rbegin(), rend(), [&key](Bindings* bind) { return bind->pressed(key); });
 }
 
 }
--- a/src/rpc/parse_commands.cc
+++ b/src/rpc/parse_commands.cc
@@ -52,17 +52,13 @@ CommandMap commands;
 XmlRpc     xmlrpc;
 ExecFile   execFile;
 
-struct command_map_is_space : std::unary_function<char, bool> {
-  bool operator () (char c) const {
-    return c == ' ' || c == '\t';
-  }
-};
+inline bool command_map_is_space(char c) {
+  return c == ' ' || c == '\t';
+}
 
-struct command_map_is_newline : std::unary_function<char, bool> {
-  bool operator () (char c) const {
-    return c == '\n' || c == '\0' || c == ';';
-  }
-};
+inline bool command_map_is_newline(char c) {
+  return c == '\n' || c == '\0' || c == ';';
+}
 
 // Only escape eol on odd number of escape characters. We know that
 // there can't be any characters in between, so this should work for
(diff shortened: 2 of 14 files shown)


src/rpc/parse_commands.cc
```diff
@@ -52,17 +52,9 @@ CommandMap commands;
...
-    return c == ' ' || c == '\t';
-  }
-};
-
-struct command_map_is_newline : std::unary_function<char, bool> {
-  bool operator () (char c) const {
-    return c == '\n' || c == '\0' || c == ';';
-  }
-};
+bool command_map_is_space(char c) {
```
rakshasa, line 55: inline

src/rpc/parse_commands.cc
```diff
@@ -150,7 +142,7 @@ parse_command(target_type target, const char* first, const char* last) {
   // Find the last character that is part of this command, skipping
   // the whitespace at the end. This ensures us that the caller
   // doesn't need to do this nor check for junk at the end.
-  first = std::find_if(first, last, std::not1(command_map_is_space()));
+  first = std::find_if(first, last, [&](char c) { return !command_map_is_space(c); });
   
   if (first != last) {
     if (*first != '\n' && *first != ';' && *first != '\0')
```
rakshasa, line 148: This should be `!command_map_is_newline(*first)`.

src/rpc/parse_commands.cc
stickz: Applied inline tag to bools.

src/rpc/parse_commands.cc
stickz: Changed to `!command_map_is_newline`.

src/input/manager.cc
```diff
@@ -64,7 +64,7 @@ Manager::pressed(int key) {
   if (m_textInput != NULL)
     m_textInput->pressed(key);
   else
-    std::find_if(rbegin(), rend(), std::bind2nd(std::mem_fun(&Bindings::pressed), key));
+    std::for_each(rbegin(), rend(), [&key](Bindings* bind) { bind->pressed(key); });
```
rakshasa, line 67: These do not do the same thing, `find_if` stops when true.

rakshasa:
```
(lldb) target create "/prefix/bin/rtorrent"
Current executable set to '/prefix/bin/rtorrent' (x86_64).
(lldb) settings set -- target.run-args  "-o" "system.daemon.set=true,import=/run/self/rc"
(lldb) settings set target.disable-aslr false
(lldb) run
rtorrent: Junk at end of input.
Process 106 exited with status = 255 (0x000000ff)
```

Getting this error.

src/input/manager.cc
stickz: Reverted to `find_if`.

stickz:
> ```
> (lldb) target create "/prefix/bin/rtorrent"
> Current executable set to '/prefix/bin/rtorrent' (x86_64).
> (lldb) settings set -- target.run-args  "-o" "system.daemon.set=true,import=/run/self/rc"
> (lldb) settings set target.disable-aslr false
> (lldb) run
> rtorrent: Junk at end of input.
> Process 106 exited with status = 255 (0x000000ff)
> ```
> 
> Getting this error.

All fixed.

```c
  if (first != last) {
    if (!command_map_is_newline(*first))
      throw torrent::input_error("Junk at end of input.");


    first++;
  }
```

rakshasa:
It needs to check all from first to last.

stickz:
> It needs to check all from first to last.

This does the same thing as before the cleanup in `src/rpc/parse_commands.cc`. If we want to change the behavior of
`parse_command(target_type target, const char* first, const char* last)`, I think we should create a new pull request. This pull request just cleans up the code to comply with the `c++17` standard. It doesn't change the algorithm of the function.

PR #1389, opened 2025-01-22
Add optional gzip compression

kannibalox:
This introduces a single new value command `network.gzip_response_min_size` for determining if the response should be gzipped (I'm very open to better names for that variable). It'll additionally require the SCGI client to announce support for gzip responses in the `ACCEPT_ENCODING` header. Since this is of dubious benefit to most use cases, it's disabled by default via setting it to a value of less than 0. This also technically makes zlib a new dependency for rtorrent, but since it's already a hard dependency for libtorrent that didn't seem unreasonable.

A fun little interaction is that calling `network.gzip_response_min_size` over RPC impacts the response immediately.

### DIFF
--- a/src/rpc/scgi_task.cc
+++ b/src/rpc/scgi_task.cc
@@ -1,4 +1,3 @@
-#include "rpc/scgi_task.h"
 #include "config.h"
 
 #include <cstdio>
@@ -13,6 +12,7 @@
 
 #include "utils/socket_fd.h"
 
+#include "scgi_task.h"
 #include "control.h"
 #include "globals.h"
 #include "scgi.h"


rakshasa:
`network.scgi.use_gzip` and `network.scgi.gzip.min_size`

src/rpc/scgi_task.cc
```diff
@@ -1,14 +1,16 @@
 #include "config.h"
 
+#include "rpc/parse_commands.h"
```
rakshasa, line 3: Do not use rpc commands here.

Instead the min size setting should be a member variable of SCgiTask.

src/rpc/scgi_task.cc
```diff
@@ -212,12 +219,14 @@ SCgiTask::event_read() {
 
 void
 SCgiTask::event_write() {
+  int bytes = -1;
```
rakshasa, line 222: This is not needed.

src/rpc/scgi_task.cc
```diff
@@ -239,36 +248,99 @@ SCgiTask::event_error() {
   close();
 }
 
+// Convenience function similar to zlib's compress(), but uses a gzip
+// header. Returns an empty string on error, as even a compressed
+// empty string will have data.
+std::string
+gzip_compress(const char* buffer, uint32_t length) {
```
rakshasa, line 255: Move to `src/utils`.

src/rpc/scgi_task.cc
```diff
@@ -239,36 +248,99 @@ SCgiTask::event_error() {
...
+    compressed_buffer = gzip_compress(buffer, length);
+    if (compressed_buffer != "") {
+      header += "Content-Encoding: gzip\r\n";
+      length = compressed_buffer.size();
+    } else {
+      // Fall back to uncompressed response if compression fails
+      should_compress = false;
+    }
+  }
+  header += "\r\n";
```
rakshasa, line 327: This use of std::strings causes unnecessary copying of buffers.

Rewrite both the gzip compressor and this to write directly to m_buffer, pass e.g. a lambda function to gzip_compressor that does the writing.

kannibalox: This causes a chicken-and-egg problem where we need to know the Content-Length to figure out where the write cursor in m_buffer should start from (given that there's a more than decent chance we might drop a digit in the size string in the course of compression), but we don't know the length until the compression itself is complete.

The only way I can think to work around that is to first write the compressed output directly a little ways into the m_buffer and `std::memmove` it back to the correct position after the headers have been written. I'll go in that direction unless you tell me otherwise.

rakshasa: Don't see why you need to do it in such a convoluted way, just pass a lambda function that does the writing and have it resize `m_buffer` if more is needed.

You can change `m_buffer` to std::vector to make it cleaner.

src/rpc/scgi_task.cc
```diff
@@ -1,14 +1,15 @@
+#include "rpc/scgi_task.h"
 #include "config.h"
```
rakshasa, line 2: `config.h` should always be the first header included.

src/rpc/scgi_task.cc
kannibalox: Moved to a static member var

src/rpc/scgi_task.cc
kannibalox: This was changed to be a member function, if it becomes generic again I'll split it out to src/utils/.

src/rpc/scgi_task.cc
```diff
@@ -239,36 +250,102 @@ SCgiTask::event_error() {
   close();
 }
 
+// On failure, returns false and the buffer is left in an
+// indeterminate state, but m_position and m_bufferSize remain the
+// same.
+bool
+SCgiTask::gzip_compress_response(const char* buffer, uint32_t length, std::string_view header_template) {
```
rakshasa, line 257: This is a bit too ugly a solution, not what I had in mind.

Add to utils a function like this:

```
gzip_compress(const char* buffer, uint32_t length, std::function<bool(char*,uint32_t,uint32_t)>)
```

The lambda function will write from m_buffer+max_header_size (where max_header_size is calculated from the current header template plus max length of the printed response size).

It will receive deflateBound value so it can resize on the first call.

Once done writing, copy the bytes from the header and response size to the buffer so they end at the start of the written gzip'ed data.

Don't use snprintf for a whole template, instead the header should be two static const strings, and you copy them and the response size.

src/rpc/scgi_task.cc
```diff
@@ -239,36 +250,102 @@ SCgiTask::event_error() {
...
   }
 
-  lt_log_print_dump(torrent::LOG_RPC_DUMP, m_buffer, m_bufferSize, "scgi", "RPC write.", 0);
+  lt_log_print_dump(torrent::LOG_RPC_DUMP, buffer, length, "scgi", "RPC write.", 0);
+
+  // Compress the response if possible
+  if (m_client_accepts_compressed_response &&
+      gzip_enabled() &&
+      length > gzip_min_size() &&
+      gzip_compress_response(buffer, length, header)) {
```
rakshasa, line 330: Don't like how a failure to compress falls back to plaintext, this should only happen if there's a serious bug so fail it completely.

src/rpc/scgi_task.cc
```diff
@@ -239,36 +250,102 @@ SCgiTask::event_error() {
...
+
+  // Need to cast due to a bug in MacOSX gcc-4.0.1.
+  if (length + header_size > std::max(m_buffer_size, (unsigned int)default_buffer_size))
+    realloc_buffer(length + header_size, NULL, 0);
+
+  m_position   = m_buffer;
+  m_buffer_size = length + header_size;
+
+  snprintf(m_buffer, m_buffer_size, header.c_str(), length);
+  std::memcpy(m_buffer + header_size, buffer, length);
```
rakshasa, line 348: If we have a compressed path in a separate function, also put the plaintext in one.

rakshasa:
@kannibalox

See [kannibalox-feature/gzip-response](https://github.com/rakshasa/rtorrent/tree/kannibalox-feature/gzip-response) for an improved implementation that doesn't do unnecessary copying of the gzip'ed response.

It's untested and likely has some bugs.

rakshasa:
`feature/gzip-scgi` (not yet done / tested properly) is the way you avoid the unnecessary copying by writing the header after compression. (with enough room for any size of content-length string)

PR #1431, opened 2025-03-29
Fix an FTBFS under -Werror=return-type

jengelh:
Some Linux distributions globally build with -Werror=return-type, to catch garbage returns from functions. process() could be called with a garbage RPCType value, in which case it would run off the end of the function since there is no `default:` case, and return new garbage.

```
rpc/rpc_manager.cc: In member function ‘bool rpc::RpcManager::process(RPCType, const char*, uint32_t, slot_response_callback)’:
rpc/rpc_manager.cc:119:1: error: control reaches end of non-void function [-Werror=return-type]
  119 | }
rpc/rpc_manager.cc: In member function ‘bool rpc::RpcManager::is_type_enabled(RPCType) const’:
rpc/rpc_manager.cc:143:1: error: control reaches end of non-void function [-Werror=return-type]
  143 | }
```

### DIFF
--- a/src/rpc/rpc_manager.cc
+++ b/src/rpc/rpc_manager.cc
@@ -115,6 +115,8 @@ RpcManager::process(RPCType type, const char* in_buffer, uint32_t length, slot_r
       return callback(response.c_str(), response.size());
     }
   }
+  default:
+    return false;
   }
 }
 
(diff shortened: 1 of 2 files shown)


rakshasa:
Why did you add 'class' to the enum?

Also, we should throw internal_error on those switch defaults.

rakshasa:
Fixed in fix/release-issues.

jengelh:
[enum class](https://en.cppreference.com/w/cpp/language/enum) is a somewhat stricter variant of enum, and since all uses of the enumerator already were scoped, I spotted an opportunity. Though, it does not change the warning. gcc still assumes there could be some external place that could potentially do process(static_cast<RPCType>(12345), ...).

PR #1465, opened 2025-05-03
Add basic VIM key navigation

fredriklindell:
Adds basic VIM key navigation (h, j, k, l)

* h -> KEY_LEFT
* j -> KEY_DOWN
* k -> KEY_UP
* l -> KEY_RIGHT

Rebind download list DISPLAY_LOG to 'L'.

### DIFF
--- a/src/ui/download_list.cc
+++ b/src/ui/download_list.cc
@@ -363,11 +363,13 @@ DownloadList::setup_keys() {
 
   m_uiArray[DISPLAY_LOG]->bindings()[KEY_LEFT] =
     m_uiArray[DISPLAY_LOG]->bindings()['B' - '@'] =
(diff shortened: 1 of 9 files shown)


rakshasa:
If this gets added it should be a config option that selects either emacs or vim key bindings, not both.

fredriklindell:
@rakshasa I'd like to take on the task of implementing the navigation mode config.

Do you already have something in your mind to share or should I upload a PR that we discuss around?

I noticed this line: [  // Move bindings into a function that defines default bindings.](https://github.com/rakshasa/rtorrent/blob/6d888c008d9b32f78c43c604da377650d411f944/src/ui/element_text.cc#L56) which makes me think that you maybe already have an idea on where to put stuff :)

Thanks in advance!

rakshasa:
That's an old comment and I'm not planning on doing anything atm, anything that improves the code is welcome.

PR #1466, opened 2025-05-03
Align compact display download header labels

fredriklindell:
Align compact display download header labels

### DIFF
--- a/src/display/utils.cc
+++ b/src/display/utils.cc
@@ -184,7 +184,7 @@ print_download_status(char* first, char* last, core::Download* d) {
 char*
 print_download_column_compact(char* first, char* last) {
   first = print_buffer(first, last, " %-64.64s", "Name");
-  first = print_buffer(first, last, "| Status | Downloaded | Size       | Done | Up Rate   | Down Rate | Uploaded   |  ETA      | Ratio| Misc ");
+  first = print_buffer(first, last, "| Status | Downloaded |    Size    | Done |  Up Rate  | Down Rate |  Uploaded  |    ETA    | Ratio | Misc ");
 
   if (first > last)
     throw torrent::internal_error("print_download_column_compact(...) wrote past end of the buffer.");
@@ -200,7 +200,7 @@ print_download_info_compact(char* first, char* last, core::Download* d) {
   if (!d->download()->info()->is_open())
     first = print_buffer(first, last, " CLOSED ");
   else if (!d->download()->info()->is_active())
-    first = print_buffer(first, last, " OPEN   ");
+    first = print_buffer(first, last, "  OPEN  ");
   else
     first = print_buffer(first, last, "        ");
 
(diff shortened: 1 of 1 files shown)


src/display/utils.cc
```diff
@@ -62,9 +62,9 @@ print_hhmmss_local(char* first, char* last, time_t t) {
 char*
 print_ddhhmm(char* first, char* last, time_t t) {
   if (t / (24 * 3600) < 100)
-    return print_buffer(first, last, "%2id %2i:%02i", (int)t / (24 * 3600), ((int)t / 3600) % 24, ((int)t / 60) % 60);
+    return print_buffer(first, last, "%2id %2i:%02i ", (int)t / (24 * 3600), ((int)t / 3600) % 24, ((int)t / 60) % 60);
   else
-    return print_buffer(first, last, "--d --:--");
+    return print_buffer(first, last, "--d --:-- ");
```
rakshasa, line 67: This is the wrong place to do the align, do it instead from where this function is called.

fredriklindell: Sure, do you mean something like this:

```
char*
print_download_time_left(char* first, char* last, core::Download* d) {
  uint32_t rate = d->info()->down_rate()->rate();

  if (rate < 512)
    return print_buffer(first, last, "--d --:-- ");
  
  time_t remaining = (d->download()->file_list()->size_bytes() - d->download()->bytes_done()) / (rate & ~(uint32_t)(512 - 1));

  if (remaining / (24 * 3600) < 100)
    return print_ddhhmm(first, last, remaining);
  else
    return print_buffer(first, last, "--d --:-- ");
}
```

And:

```
char*
print_ddhhmm(char* first, char* last, time_t t) {
  return print_buffer(first, last, "%2id %2i:%02i ", (int)t / (24 * 3600), ((int)t / 3600) % 24, ((int)t / 60) % 60);
}
```

rakshasa: No, you should not start ignoring large values.

And you shouldn't be adding the whitespace after the ddhhmm, instead add it in front.

The 'remaining' check can be used to decide if you want to add whitespace before `print_ddhhmm` or not in `print_download_time_left`.

fredriklindell: Not sure that I agree on where to add the extra space, looking at the screenshot I recommend to add it at the end

2025-05-07-072528-selection

I'll look into the logic of the remaining check

rakshasa: Then also add the space after.

src/display/utils.cc
```diff
@@ -223,9 +223,9 @@ print_download_info_compact(char* first, char* last, core::Download* d) {
   if (d->download()->info()->is_active() && !d->is_done())
     first = print_download_time_left(first, last, d);
   else
-    first = print_buffer(first, last, "         ");
+    first = print_buffer(first, last, "          ");
 
-  first = print_buffer(first, last, "| %4.2f ", (double)rpc::call_command_value("d.ratio", rpc::make_target(d)) / 1000.0);
+  first = print_buffer(first, last, "|  %4.2f ", (double)rpc::call_command_value("d.ratio", rpc::make_target(d)) / 1000.0);
```
rakshasa, line 228: Increase to 5 here.

src/display/utils.cc
fredriklindell: I moved the logic to `print_download_info_compact`, that way it will not inflict any changes to the full layout.

Was that the correct place to add the logic?

PR #1473, opened 2025-05-05
Removed/replaced deprecated commands (execute/schedule/schedule_remove) and removed global lock in ExecFile.

rakshasa:


rakshasa:
The execute/schedule/schedule_remove commands now require a target, so it might break any users of RPC that haven't updated to the new command style. (after 15 years)

PR #1484, opened 2025-05-19
Add emacs and vi keymap config logic

fredriklindell:
* Add emacs and vi keymap config logic

### DIFF
--- a/src/command_ui.cc
+++ b/src/command_ui.cc
@@ -824,6 +824,9 @@ initialize_command_ui() {
   CMD2_ANY_LIST  ("ui.status.throttle.up.set",   std::bind(&cmd_status_throttle_names, true, std::placeholders::_2));
   CMD2_ANY_LIST  ("ui.status.throttle.down.set", std::bind(&cmd_status_throttle_names, false, std::placeholders::_2));
 
+  CMD2_ANY         ("ui.keymap.style",     std::bind(&ui::Root::keymap_style, control->ui()));
+  CMD2_ANY_STRING_V("ui.keymap.style.set", std::bind(&ui::Root::set_keymap_style, control->ui(), std::placeholders::_2));
+
   // TODO: Add 'option_string' for rtorrent-specific options.
   CMD2_VAR_STRING("ui.torrent_list.layout", "full");
 
--- a/src/ui/download_list.cc
+++ b/src/ui/download_list.cc
@@ -360,12 +360,12 @@ DownloadList::setup_keys() {
   m_bindings['F']           = std::bind(&DownloadList::receive_view_input, this, INPUT_FILTER);
 
   m_uiArray[DISPLAY_LOG]->bindings()[KEY_LEFT] =
-    m_uiArray[DISPLAY_LOG]->bindings()['B' - '@'] =
+    m_uiArray[DISPLAY_LOG]->bindings()[control->ui()->navigation_key(RT_KEY_LEFT)] =
     m_uiArray[DISPLAY_LOG]->bindings()[' '] = std::bind(&DownloadList::activate_display, this, DISPLAY_DOWNLOAD_LIST);
 
   m_uiArray[DISPLAY_DOWNLOAD_LIST]->bindings()[KEY_RIGHT] =
-    m_uiArray[DISPLAY_DOWNLOAD_LIST]->bindings()['F' - '@'] = std::bind(&DownloadList::activate_display, this, DISPLAY_DOWNLOAD);
-  m_uiArray[DISPLAY_DOWNLOAD_LIST]->bindings()['l'] = std::bind(&DownloadList::activate_display, this, DISPLAY_LOG);
+    m_uiArray[DISPLAY_DOWNLOAD_LIST]->bindings()[control->ui()->navigation_key(RT_KEY_RIGHT)] = std::bind(&DownloadList::activate_display, this, DISPLAY_DOWNLOAD);
+  m_uiArray[DISPLAY_DOWNLOAD_LIST]->bindings()[control->ui()->navigation_key(RT_KEY_DISPLAY_LOG)] = std::bind(&DownloadList::activate_display, this, DISPLAY_LOG);
 }
 
 }
--- a/src/ui/root.cc
+++ b/src/ui/root.cc
@@ -27,6 +27,34 @@
 
 namespace ui {
 
+static const int emacs_keymap[RT_KEYMAP_MAX] = {
+  'B' - '@',
+  'F' - '@',
+  'P' - '@',
+  'N' - '@',
+  'l',
+  'k',
+  'U' - '@',
+  'H' - '@',
+  'A' - '@',
+  'E' - '@',
+  'L'
+};
+
+static const int vi_keymap[RT_KEYMAP_MAX] = {
+  'h',
+  'l',
+  'k',
+  'j',
+  'L',
+  'K',
+  'B' - '@',
+  'D' - '@',
+  'H',
+  'G',
+  'L' - '@'
+};
+
 Root::Root() {
   // Initialise prefilled m_input_history and m_input_history_pointers objects.
   for (int type = ui::DownloadList::INPUT_LOAD_DEFAULT; type != ui::DownloadList::INPUT_EOI; type++) {
@@ -34,6 +62,9 @@ Root::Root() {
     m_input_history_pointers.insert( std::make_pair(type, 0) );
   }
 
+  // set default keymap to emacs
+  m_keymap_style = "emacs";
+  m_keymap = emacs_keymap;
 }
 
 void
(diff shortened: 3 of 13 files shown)


src/command_ui.cc
```diff
@@ -872,6 +872,8 @@ initialize_command_ui() {
   CMD2_ANY_LIST ("elapsed.less",         std::bind(&apply_elapsed_less, std::placeholders::_2));
   CMD2_ANY_LIST ("elapsed.greater",      std::bind(&apply_elapsed_greater, std::placeholders::_2));
 
+  CMD2_VAR_STRING("navigation.mode", "emacs");
```
rakshasa, line 875: `ui.keymap.style` and move above.

src/ui/download_list.cc
```diff
@@ -361,13 +362,13 @@ DownloadList::setup_keys() {
...
   m_bindings['F']           = std::bind(&DownloadList::receive_view_input, this, INPUT_FILTER);
 
-  m_uiArray[DISPLAY_LOG]->bindings()[KEY_LEFT] =
+  m_uiArray[DISPLAY_LOG]->bindings()[control->ui()->get_keymap(input::RT_KEY_LEFT)] =
     m_uiArray[DISPLAY_LOG]->bindings()['B' - '@'] =
     m_uiArray[DISPLAY_LOG]->bindings()[' '] = std::bind(&DownloadList::activate_display, this, DISPLAY_DOWNLOAD_LIST);
 
-  m_uiArray[DISPLAY_DOWNLOAD_LIST]->bindings()[KEY_RIGHT] =
+  m_uiArray[DISPLAY_DOWNLOAD_LIST]->bindings()[control->ui()->get_keymap(input::RT_KEY_RIGHT)] =
     m_uiArray[DISPLAY_DOWNLOAD_LIST]->bindings()['F' - '@'] = std::bind(&DownloadList::activate_display, this, DISPLAY_DOWNLOAD);
```
rakshasa, line 370: ^F and ^B are the emacs keys, so should not active.

KEY_LEFT/RIGHT are not specific to emacs, so should remain.

src/ui/root.cc
```diff
@@ -451,4 +452,14 @@ Root::clear_input_history() {
   }
 }
 
+const int
+Root::get_keymap(int key) {
+  const std::string& navigationMode = rpc::call_command_string("navigation.mode");
```
rakshasa, line 457: Instead of calling command each time, add an 'foo' and 'foo_set' for std::string to Root, and change the command macro to the ones that set strings.

fredriklindell: Do you have any reference in the code as an example? Thanks in advance!

src/command_ui.cc
fredriklindell: Do you mean remove above and instead use the `ui.keymap.style` in the config file?

src/ui/root.cc
fredriklindell: Do you mean like this:

```
  const std::string layout_name = rpc::call_command_string("ui.torrent_list.layout");
```

rakshasa:
```
  CMD2_ANY       ("ui.current_view",       std::bind(&cmd_ui_current_view));
  CMD2_ANY_STRING("ui.current_view.set",   std::bind(&cmd_ui_set_view, std::placeholders::_2));

  CMD2_ANY        ("ui.input.history.size",     std::bind(&ui::Root::get_input_history_size, control->ui()));
  CMD2_ANY_VALUE_V("ui.input.history.size.set", std::bind(&ui::Root::set_input_history_size, control->ui(), std::placeholders::_2));
```

src/ui/download_list.cc
fredriklindell: What does the `- '@'` do?

src/Makefile.am
```diff
@@ -88,6 +88,7 @@ libsub_root_a_SOURCES = \
 	input/input_event.h \
 	input/manager.cc \
 	input/manager.h \
+	input/navigation_keymap.h \
```
fredriklindell, line 91: @rakshasa I have made the changes that you proposed, thanks for the review.

I'm not sure if the naming or the location of the navigation-keymap.h file is any good, do you have a better solution?

Thanks in advance

src/command_ui.cc
rakshasa: No, i just want the name changed.

fredriklindell: Alright, it doesn't seem to be necessary after the 'foo' and 'foo_set' change you suggested :)

src/ui/root.cc
```diff
@@ -451,4 +452,18 @@ Root::clear_input_history() {
...
+}
+
+
+const int
+Root::get_keymap(int key) {
+  if (strcasecmp(m_keymap_style.c_str(), "vi") == 0)
+    return input::vi_keymap[key];
+  else
+    return input::emacs_keymap[key];
+}
```
rakshasa, line 467: Move these and vi/emacs_keymap to root.cc, no need for `navigation_keymap.*`.

You should have if statements for `vi`, `emacs` and empty string, with error if none is found.

Don't use `c_str()`, just straight case-sensetive std::string comparison. We are case sensitive everywhere else.

fredriklindell: I have (re)moved the `navigation_keymap.h` to `root.h` and tried fixing the `c_str()` removal, but I can't figure it out. The `CMD_ANT_STRING_V` gives me a `std:string` but the `"vi"` is of type `char*` and it does not seem to be possible to compare the two without the `c_str()`. I looked around the code base and it seems like this way of doing this is used in a lot of places.

Do you have any suggestion? Should I use another `CMD2_*` function?

Thanks in advance!

rakshasa: `m_keymap_style == "vi"` should work.

src/ui/root.h
```diff
@@ -23,6 +23,49 @@ namespace input {
...
+  'k',
+  'j',
+  'L',
+  'K',
+  'B' - '@',
+  'D' - '@',
+  'H',
+  'G',
+  'L' - '@'
+};
```
rakshasa, line 67: These two should be in root.cc.

fredriklindell: Thanks, done!

New push up with changes

src/ui/root.cc
fredriklindell: Many thanks that worked :+1:

fredriklindell (commented):
@rakshasa How do I solve the static code analysis, I can't figure out what it complains on.

Anything else I need to fix in this PR?

Thanks in advance!

rakshasa:
Likely need to merge latest master, you can see the error messages.

fredriklindell:
@rakshasa Thanks, the rebase solved the static analysis.

Anything else that I need to do?

src/ui/root.h
```diff
@@ -80,6 +95,10 @@ class Root {
   void                save_input_history();
   void                clear_input_history();
 
+  const std::string&  get_keymap_style()                      { return m_keymap_style; }
+  void                set_keymap_style(const std::string& style);
+  const int           get_keymap(int key);
```
rakshasa, line 100: `get_keymap_style` should be `keymap_style`.

fredriklindell: Updated!

src/ui/root.cc
```diff
@@ -451,4 +479,20 @@ Root::clear_input_history() {
   }
 }
 
+void
+Root::set_keymap_style(const std::string& style) {
+  m_keymap_style = style;
```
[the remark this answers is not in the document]
fredriklindell: Great catch :+1:

While changing this I had to remove the default m_keymap_style `emacs` value. And now the `ui.keymap.style` needs to be set to either `emacs` or `vi` in the `rtorrent.rc` file.

@rakshasa Please check if the change was in line with what you thought.

Thanks in advance!

src/ui/root.cc
```diff
@@ -451,4 +479,26 @@ Root::clear_input_history() {
...
+    throw torrent::internal_error("Root::set_keymap_style() -> ui.keymap.style is configured with unknown keymap style: " + m_keymap_style);
+  }
+}
+
+const int
+Root::get_keymap(int key) {
+
+  if (m_keymap == NULL)
+    throw torrent::internal_error("Root::get_keymap() -> ui.keymap.style is not set or is invalid, please set ui.keymap.style in configuration file.");
+
```
rakshasa, line 500: Remove this check.

src/ui/root.h
rakshasa: `get_keymap` should be `navigation_key(NavigationKeymap key)`.

src/ui/root.cc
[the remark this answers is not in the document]
fredriklindell: Done!

mmmsalmon:
`^d` to stop/remove a torrent doesn't seem to work when `ui.keymap.style.set = "vi"` is set.
Is there an alternative, or is this a bug?

rakshasa:
Looks like it needs to be remapped, or fixed in some other way.

Seems it is mapped to RT_KEY_NPAGE, however I don't use vi so don't know what is the right keys.

mmmsalmon:
Could be set to `x` (delete character in vi) or `dd` (delete line in vi).
I could make a PR but I have time only on weekends.

PR #1505, opened 2025-05-31
android termux phone view small screen file list

john-peterson:
you can pin this for phone users desktop users are going to panic at
this. like windows eight. there is a hidden secret that few people know about. desktops are
meaningless phones can do everything . but don't tell anyone  hush

 # to-do

I would also like F filter to apply in the file list if there are many
files. for emulator like Mame there can be thousands files and I only
need one


rakshasa:
Closing this as it's not going to be merged.

PR #1528, opened 2025-06-17
get nlohmann json from pkgconfig

neheb:
Many distros already include it as a package.


rakshasa:
I want less external dependencies, not more.

So this goes completely counter to this preference.

rakshasa:
I want less external dependencies, not more.

So this goes completely counter to this preference.

PR #1529, opened 2025-06-17
small fixes

neheb:


src/display/attributes.h
```diff
@@ -4,15 +4,15 @@
...
+#elif __has_include(<ncursesw.h>)
 #include <ncursesw.h>
-#elif defined(HAVE_NCURSES_CURSES_H)
+#elif __has_include(<ncurses/curses.h>)
 #include <ncurses/curses.h>
-#elif defined(HAVE_NCURSES_H)
+#elif __has_include(<ncurses.h>)
 #include <ncurses.h>
-#elif defined(HAVE_CURSES_H)
+#elif __has_include(<curses.h>)
```
rakshasa, line 15: This is wrong as it removes the ability to control which header to use.

src/command_local.cc
```diff
@@ -204,7 +204,7 @@ initialize_command_local() {
   CMD2_ANY         ("system.pid",      std::bind(&getpid));
 
   CMD2_VAR_C_STRING("system.api_version",           (int64_t)API_VERSION);
-  CMD2_VAR_C_STRING("system.client_version",        PACKAGE_VERSION);
+  CMD2_VAR_C_STRING("system.client_version",        VERSION);
```
rakshasa, line 207: Why is this better?

rakshasa: I much prefer PACKAGE_VERSION as it is more explicit.

neheb: Done.

src/display/attributes.h
rakshasa:
```
#     (no options)                           - NcursesW, Ncurses, Curses
#     --with-ncurses     --with-ncursesw     - NcursesW only [*]
#     --without-ncurses  --with-ncursesw     - NcursesW only [*]
#                        --with-ncursesw     - NcursesW only [*]
#     --with-ncurses     --without-ncursesw  - Ncurses only [*]
#     --with-ncurses                         - NcursesW, Ncurses [**]
#     --without-ncurses  --without-ncursesw  - Curses only
#                        --without-ncursesw  - Ncurses, Curses
#     --without-ncurses                      - NcursesW, Curses
#
```

neheb: these are mutually exclusive, no?

The code uses only one of these. It wants curses, no matter the source.

rakshasa: Also, the current code is how the curses autoconf script recommends one do it.

neheb: don't agree with this but removed.

PR #1536, opened 2025-06-20
command_ui: use templates instead of function

neheb:
Less overhead.


neheb:
I tested size increase

Before:
```
ls -l a/src/rtorrent;strip a/src/rtorrent;ls -l a/src/rtorrent
-rwxr-xr-x. 1 mangix mangix 38184352 Jun 20 16:25 a/src/rtorrent*
-rwxr-xr-x. 1 mangix mangix 3731928 Jun 20 16:26 a/src/rtorrent*
```

After:
```
-rwxr-xr-x. 1 mangix mangix 37944200 Jun 20 16:27 a/src/rtorrent*
-rwxr-xr-x. 1 mangix mangix 3711448 Jun 20 16:27 a/src/rtorrent*
```

no idea why.

PR #1538, opened 2025-06-20
some more get() removals

neheb:


neheb:
changed those binds to lambdas to avoid std::ref.

src/command_local.cc
```diff
@@ -211,9 +211,8 @@ initialize_command_local() {
   CMD2_VAR_VALUE   ("system.file.split_size",       -1);
   CMD2_VAR_STRING  ("system.file.split_suffix",     ".part");
 
-  CMD2_ANY         ("system.file_status_cache.size",   std::bind(&utils::FileStatusCache::size,
-                                                                 (utils::FileStatusCache::base_type*)control->core()->file_status_cache()));
-  CMD2_ANY_V       ("system.file_status_cache.prune",  std::bind(&utils::FileStatusCache::prune, control->core()->file_status_cache()));
+  CMD2_ANY         ("system.file_status_cache.size",   [](auto...) { return reinterpret_cast<utils::FileStatusCache::base_type*>(control->core()->file_status_cache().get())->size(); });
```
neheb, line 214: why is this cast here?

neheb: the class already has using base_type::size;

this is not needed.

src/core/manager.h
```diff
@@ -36,9 +36,9 @@ class Manager {
   Manager();
   ~Manager();
 
-  DownloadList*       download_list()                     { return m_download_list.get(); }
-  DownloadStore*      download_store()                    { return m_download_store.get(); }
-  FileStatusCache*    file_status_cache()                 { return m_file_status_cache.get(); }
+  const auto&         download_list() const               { return m_download_list; }
+  const auto&         download_store() const              { return m_download_store; }
+  const auto&         file_status_cache() const           { return m_file_status_cache; }
```
rakshasa, line 41: This changes the behavior of the code, not sure I want this.

I'd rather use the old pointers in these cases.

neheb: Does it really matter? It’s still a non const pointer.

PR #1539, opened 2025-06-21
regex: remove std::inheritance

neheb:
Simpler than std::list.

Remove std::function. No need for it.

### DIFF
--- a/rak/regex.h
+++ b/rak/regex.h
@@ -50,7 +50,7 @@
 
 namespace rak {
 
-class regex : public std::function<bool (std::string)> {
+class regex {
 public:
   regex() {}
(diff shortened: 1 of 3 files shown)


rak/regex.h
```diff
@@ -93,17 +90,17 @@ regex::operator () (const std::string& text) const {
 
       // Push to the back so that '*' will match zero length strings.
       if (m_pattern[next] == '*')
-	paths.push_back(next);
+	paths.insert(next);
```
rakshasa, line 93: Note the comment, these changes do not produce the same result.

neheb:
@rakshasa removed.

PR #1546, opened 2025-07-19
Optimize load_session_torrents code path

jesec:
| Torrents / Time to Load (seconds) | w/o this PR | w/ this PR |
| --------------------------------- | ----------- | ---------- |
| 100                               | 0.0988322   | 0.100783   |
| 500                               | 0.316298    | 0.2778484  |
| 1000                              | 0.5230036   | 0.3384636  |
| 2000                              | 1.1452586   | 0.488410   |
| 5000                              | 5.4459174   | 0.8843604  |
| 10000                             | 24.0999498  | 1.574791   |
| 15000                             | 62.0685972  | 2.3090436  |
| 20000                             | 122.2196932 | 3.118153   |
| 25000                             | 194.3020272 | 3.9291368  |
| 30000                             | 301.594044  | 4.7896626  |

---

image

### DIFF
--- a/src/core/download_factory.cc
+++ b/src/core/download_factory.cc
@@ -297,11 +297,11 @@ DownloadFactory::receive_success() {
     for (const auto& command : m_commands)
       rpc::parse_command_multiple_std(command, rpc::make_target(download));
 
-    if (m_manager->download_list()->find(infohash) == m_manager->download_list()->end())
-      throw torrent::input_error("The newly created download was removed.");
-
-    if (!m_session)
-       rpc::call_command("d.state.set", (int64_t)m_start, rpc::make_target(download));
+    if (!m_session) {
+      if (m_manager->download_list()->find(infohash) == m_manager->download_list()->end())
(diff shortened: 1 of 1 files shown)


neheb:
impressive.

src/main.cc
```diff
@@ -302,16 +301,24 @@ main(int argc, char** argv) {
...
+    // Only enable view sorting when not in daemon mode
+    if (!rpc::call_command_value("system.daemon")) {
+      rpc::parse_command_multiple
+        (rpc::make_target(),
+         "view.sort_new     = name,((less,((d.name))))\n"
+         "view.sort_current = name,((less,((d.name))))\n"
+         "schedule2 = view.main,10,10,((view.sort,main,20))\n"
+         "schedule2 = view.name,10,10,((view.sort,name,20))\n"
+        );
+    }
```
rakshasa, line 320: Daemon mode still uses views, so the name view should be sorted.

However this could be changed to be added on `event.system.startup_done`, which likely should also be used to enable scgi if opened.

PR #1549, opened 2025-08-03
vi/emacs keybind for stopping torrent from download list

mmmsalmon:
Related to PR #1484

I'm not familiar with C++ and I wasn't able to test this since I couldn't get the compiler to find the headers from libtorrent. In theory this PR should make `x` in vi mode the same as `^d`. `C-d` happens to be delete character in emacs, which is the same as the current ncurses binding.


rakshasa:
Fixed PR: #1550

fredriklindell:
Nice catch and fix! I missed those unicode bindings. Thanks for reporting and fix! :+1:

Issue #1559, opened 2025-08-30
Apt Package and Debian repositories out of date


beadon:
The latest rtorrent version on the debian package manager is 0.9.8.  The package maintainer for debian is ghostbar@debian.org .  The homepage listed in the debian package notes is https://rakshasa.github.io/rtorrent/ , but it seems that the releases is a bit more recent (https://github.com/rakshasa/rtorrent/releases/).

Can someone get in touch with the package maintainer for debian / apt so that rtorrent may see the latest stable versions ?

Or .. otherwise maybe have a build-script auto-submit it ?

rakshasa:
Upstream is the responsibility of motived user or maintainers, so nothing I'll be touching.

asdflkasdf:
The package in the Debian repo has been out of date for **eons**.

I highly encourage everyone to contact the maintainers of the package and compel them to roll out an update. 

https://tracker.debian.org/pkg/rtorrent

beadon:
have opened a ticket with Ubuntu to start -- they're pointing to Debian maintainers.

https://answers.launchpad.net/ubuntu/+source/rtorrent/+question/822409

@rakshasa please see the comment in the open issue about version numbers.  I believe a small update to documentation will go a long way in helping people understand what the "latest" is:

> I am not sure about the versioning of rtorrent.
> https://rakshasa.github.io/rtorrent/ still lists 0.9.8 as "stable" release and does not mention 0.15.6 at all.

rakshasa:
Fixed, forgot to update that page.

beadon:
https://tracker.debian.org/pkg/rtorrent
Is currently showing that the latest version is available here ..

It seems that the Debian package is linked to this repo:
https://salsa.debian.org/debian/rtorrent

.. but  I am unclear how to get it to link to *this* GitHub repo as upstream ..

beadon:
Have opened a bug with Debian -- a maintainer has indicated that if there is no action after a period of time, I can step in as the maintainer and package it for Debian.

reference : https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1112602

There are a number of bugs stuck in the Debian queue for the package - some of which have probably been resolved.  Just letting you know @rakshasa ( https://bugs.debian.org/cgi-bin/pkgreport.cgi?repeatmerged=no&src=rtorrent ).   Package maintainer needs to tidy a bit more, or that could become something we sort if control is given.

beadon:
[...]

PR #1605, opened 2025-10-14
Add packaging status

ghost:
A bit related to: https://github.com/rakshasa/rtorrent/issues/1559

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,3 +86,29 @@ compiled if the user wishes to avoid using OpenSSL.
 * autoconf
 * autoheader
 * automake
+
+## Packaging status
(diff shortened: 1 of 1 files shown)


rakshasa:
Big images shouldn't be embedded in the readme.

ghost:
Specific distributions badges could be used instead.
https://repology.org/project/unit/badges

Which would you pick from that list?
Or do you even agree with the PR idea?

rakshasa:
If it's limited to major distros it should be ok.

ghost:
Alright.
How does this look?
https://github.com/WaffleBadness/rtorrent/blob/versionStatus/README.md#packaging-status

rakshasa:
Still way too many, probably better to just make a link to a wiki page where everything is listed.

PR #1690, opened 2026-01-12
Prevents rtorrent: caught torrent::internal_error: SessionManager::replace_save_request_unsafe() cannot replace full save requests.

PiloUnk:
Modified `SessionManager::replace_save_request_unsafe()` to return `false` instead of throwing when encountering a full save request

### DIFF
--- a/src/session/session_manager.cc
+++ b/src/session/session_manager.cc
@@ -104,8 +104,6 @@ SessionManager::save_full_download(core::Download* download) {
     storer.libtorrent_stream()
   };
 
-  bool should_schedule = false;
-
   {
     std::unique_lock<std::mutex> lock(m_mutex);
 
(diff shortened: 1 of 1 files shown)


rakshasa:
That is not the right solution afaict.

I'll need to review the code, however I think the two other streams should just be replaces as normal even if it is a full resume save. (full resume are only done as a torrent is created)

PiloUnk:
This supersedes the previous attempt which was indeed too "brutal".

The first removes the hard error in `replace_save_request_unsafe()` and instead merges incoming resume data into an already-queued full save, preventing the `cannot replace full save requests` crash. The second ensures `process_save_request()` is always scheduled after the merge, so the updated save is never silently dropped.

I've been using it for 1 month without any issue.

PR #1710, opened 2026-02-26
Add trusted/untrusted XMLRPC connection security model

xirvik:
This is a port (well, reimplementation, because there was no way to use the original code in the new rtorrent, so much has changed)  from Novik's rtorrent work.
 
## Summary

Replaces the v2 blacklist approach (static `unordered_set` checked at the RPC entry point) with a whitelist model using per-command flags checked in `call_command()`. This addresses the three problems identified in the v2 review:

1. **Nested command bypass** — v2 only checked the top-level RPC call; commands invoked via `branch`, `if`, argument expansion, etc. were unchecked. v3 enforces in `call_command()`, catching all execution paths.
2. **Blacklist fragility** — new or user-defined commands were allowed by default. v3 defaults to deny; commands must opt in with `flag_untrusted_safe`.
3. **Simpler and less fragile** — the separate blocked-command set is removed; each command's own `m_flags` field carries the safety annotation, so the check and the registration live in the same place.

### How it works

- New flag `flag_untrusted_safe = 0x400` in `command_map.h`.
- `call_command()` (both overloads) checks `!is_trusted() && !flag_untrusted_safe` and throws `untrusted_error`.
- `untrusted_error` is caught and translated to protocol-specific error codes in all three handlers: `xmlrpc_c.cc` (→ -507), `xmlrpc_tinyxml2.cc` (→ -507), `jsonrpc.cc` (→ -32601).
- Trust state is a `thread_local bool` in `RpcManager`, set per-request by the SCGI layer based on the `UNTRUSTED_CONNECTION` header.
- `_U` macro variants (`CMD2_ANY_U`, `CMD2_DL_U`, `CMD2_VAR_VALUE_U_GET`, etc.) set the flag at registration time.
- `_U_GET` variants mark the getter safe but leave the setter trusted-only — used for `network.rpc.use_xmlrpc` and similar settings where untrusted read is fine but untrusted write would be a DoS vector.
- Dynamic commands created by `group_insert()` have the flag propagated after creation.

Trusted connections are completely unaffected — `m_trusted` defaults to `true`, so the flag check is skipped and all commands execute exactly as before.

### Commits

1. **`80d88a2` — Infrastructure**: flag, `untrusted_error` exception, enforcement in `call_command()`, catch blocks in all 3 protocol handlers, SCGI trust state management, `_U` macro variants.
2. **`0dc3aa3` — Whitelist annotations**: ~120 safe commands marked across 13 source files.
3. **`47afd59` — Crash fix**: `network.rpc.use_xmlrpc` / `network.rpc.use_jsonrpc` were not marked safe but are called by `RpcManager::process()` before dispatching — threw `untrusted_error` which escaped the callback and terminated the process. Marked as `CMD2_VAR_BOOL_U`.
4. **`259f625` — Review fix**: Changed the above two to `CMD2_VAR_BOOL_U_GET` (getter safe, setter trusted-only) to prevent untrusted callers from disabling RPC transports. Removed overly broad `catch(std::exception&)` / `catch(...)` added during debugging.

## Test plan

25/25 tests passing on seedboxdev003 using the SCGI test harness (posted in a PR comment). Test sections:

1. Basic trusted XMLRPC connectivity (`system.client_version`, `system.library_version`, `system.listMethods`)
2. Safe commands work for untrusted (`system.client_version`, `d.multicall2`, `throttle.global_up.rate`)
3. Dangerous commands blocked for untrusted (`execute.throw`, `execute.capture`, `execute.nothrow`, `method.insert`, `schedule2`, `import`, `log.open_file`, `file.append`)
4. Dangerous commands allowed for trusted (`execute.capture`)
5. Monitoring commands (`network.open_files`, `throttle.max_unchoked_uploads`, etc.)
6. ruTorrent throttle commands (`throttle.global_up.max_rate`, `throttle.global_down.max_rate`)
7. RPC transport setter regression — untrusted getter succeeds, untrusted `.set` blocked (`network.rpc.use_xmlrpc`, `network.rpc.use_jsonrpc`)

Full test script will be posted as a PR comment.

### DIFF
--- a/src/command_dynamic.cc
+++ b/src/command_dynamic.cc
@@ -461,7 +461,6 @@ initialize_command_dynamic() {
 #endif
 
   rpc::rpc.mark_safe("method.use_deprecated");
-  rpc::rpc.mark_safe("method.use_deprecated.set");
   rpc::rpc.mark_safe("method.const");
   rpc::rpc.mark_safe("method.has_key");
   rpc::rpc.mark_safe("method.list_keys");


rakshasa:
This looks like it only blacklists commands at the root level, while the way rtorrent handles things you can have nested commands / expand arguments.

So every single command executed while in an unsecured connection would need to check the blacklist.

And I don't really trust blacklists, too easy to forget to update when doing changes, or when a user adds commands. It should be a whitelist.

Also a `std::set` lookup for each call is inefficient, and should instead be a flag in the rpc command object.

xirvik:
Test script:

```
#!/usr/bin/env python3
"""Test rtorrent 0.16.6 XMLRPC over SCGI — trusted and untrusted connections."""

import socket
import sys
import xml.etree.ElementTree as ET

HOST = "localhost"
PORT = 5000

def scgi_call(method, params_xml="", untrusted=False):
    """Send an XMLRPC call over SCGI. Returns (success, response_text)."""
    body = (
        '<?xml version="1.0"?>'
        '<methodCall>'
        '<methodName>{}</methodName>'
        '<params>{}</params>'
        '</methodCall>'
    ).format(method, params_xml)

    # Build SCGI headers as null-terminated key-value pairs
    headers = {
        "CONTENT_LENGTH": str(len(body)),
        "SCGI": "1",
    }
    if untrusted:
        headers["UNTRUSTED_CONNECTION"] = "1"

    header_str = "\x00".join(
        k + "\x00" + v for k, v in headers.items()
    ) + "\x00"

    request = "{}:{},{}".format(len(header_str), header_str, body)

    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    s.settimeout(10)
    try:
        s.connect((HOST, PORT))
        s.sendall(request.encode())

        response = b""
        while True:
            try:
                chunk = s.recv(4096)
                if not chunk:
                    break
                response += chunk
            except socket.timeout:
                break
    finally:
        s.close()

    # Split HTTP header from body
    parts = response.split(b"\r\n\r\n", 1)
    body_text = parts[1].decode() if len(parts) > 1 else response.decode()
    return body_text


def parse_xmlrpc_response(xml_text):
    """Parse XMLRPC response, return (is_fault, value_or_message)."""
    try:
        root = ET.fromstring(xml_text)
    except ET.ParseError as e:
        return True, f"XML parse error: {e}\nRaw: {xml_text[:500]}"

    fault = root.find(".//fault")
    if fault is not None:
        fault_string = fault.find(".//member[name='faultString']/value/string")
        if fault_string is not None:
            return True, fault_string.text
        return True, ET.tostring(fault, encoding="unicode")

    # Extract value
    value = root.find(".//params/param/value")
    if value is not None:
        # Try string, i4, i8, etc.
        for child in value:
            return False, child.text
        return False, value.text
    return False, xml_text[:200]


def test(description, method, params_xml="", untrusted=False, expect_blocked=False):
    """Run a test and print result."""
    mode = "UNTRUSTED" if untrusted else "TRUSTED"
    try:
        response = scgi_call(method, params_xml, untrusted=untrusted)
        is_fault, msg = parse_xmlrpc_response(response)

        if expect_blocked:
            if is_fault and "not allowed for untrusted" in (msg or ""):
                print(f"  PASS [{mode}] {description}")
                print(f"       Correctly blocked: {msg}")
                return True
            else:
                print(f"  FAIL [{mode}] {description}")
                print(f"       Expected blocked, got: fault={is_fault} msg={msg}")
                return False
        else:
            if is_fault:
                print(f"  FAIL [{mode}] {description}")
                print(f"       Unexpected fault: {msg}")
                return False
            else:
                print(f"  PASS [{mode}] {description}")
                print(f"       Response: {msg}")
                return True
    except Exception as e:
        print(f"  FAIL [{mode}] {description}")
        print(f"       Exception: {e}")
        return False


EMPTY_TARGET = '<param><value><string></string></value></param>'

def string_param(s):
    return f"<param><value><string>{s}</string></value></param>"


def int_param(n):
    return f"<param><value><i8>{n}</i8></value></param>"


def main():
    passed = 0
    failed = 0
    total = 0

    def run(desc, method, params="", untrusted=False, expect_blocked=False):
        nonlocal passed, failed, total
        total += 1
        if test(desc, method, params, untrusted, expect_blocked):
            passed += 1
        else:
            failed += 1

    print("=" * 60)
    print("rtorrent 0.16.6 XMLRPC test suite")
    print("=" * 60)

    # --- Section 1: Basic XMLRPC connectivity ---
    print("\n--- 1. Basic XMLRPC connectivity (trusted) ---")
    run("system.client_version", "system.client_version")
    run("system.library_version", "system.library_version")
    run("system.listMethods (returns list)", "system.listMethods")

    # --- Section 2: Safe commands work for both trusted and untrusted ---
    print("\n--- 2. Safe commands work for untrusted connections ---")
    run("system.client_version (untrusted)", "system.client_version", untrusted=True)
    run("d.multicall2 (untrusted)", "d.multicall2",
        '<param><value><string></string></value></param>'
        '<param><value><string>main</string></value></param>'
        '<param><value><string>d.name=</string></value></param>',
        untrusted=True)
    run("throttle.global_up.rate (untrusted)", "throttle.global_up.rate", untrusted=True)

    # --- Section 3: Dangerous commands BLOCKED for untrusted ---
    # v0.16.6 requires empty target string as first param for these commands
    print("\n--- 3. Dangerous commands blocked for untrusted ---")
    run("execute.throw blocked", "execute.throw",
        EMPTY_TARGET + string_param("echo") + string_param("hello"),
        untrusted=True, expect_blocked=True)
    run("execute.capture blocked", "execute.capture",
        EMPTY_TARGET + string_param("echo") + string_param("hello"),
        untrusted=True, expect_blocked=True)
    run("execute.nothrow blocked", "execute.nothrow",
        EMPTY_TARGET + string_param("echo") + string_param("hello"),
        untrusted=True, expect_blocked=True)
    run("method.insert blocked", "method.insert",
        EMPTY_TARGET + string_param("test_method") + string_param("simple") + string_param(""),
        untrusted=True, expect_blocked=True)
    run("schedule2 blocked", "schedule2",
        EMPTY_TARGET + string_param("test") + string_param("10") + string_param("10") + string_param("print=test"),
        untrusted=True, expect_blocked=True)
    run("import blocked", "import",
        EMPTY_TARGET + string_param("/tmp/nonexistent.rc"),
        untrusted=True, expect_blocked=True)
    run("log.open_file blocked", "log.open_file",
        EMPTY_TARGET + string_param("test") + string_param("/tmp/test.log"),
        untrusted=True, expect_blocked=True)
    run("file.append blocked", "file.append",
        EMPTY_TARGET + string_param("/tmp/test.txt") + string_param("data"),
        untrusted=True, expect_blocked=True)

    # --- Section 4: Dangerous commands ALLOWED for trusted ---
    print("\n--- 4. Dangerous commands allowed for trusted ---")
    run("execute.capture (trusted)", "execute.capture",
        EMPTY_TARGET + string_param("echo") + string_param("hello_from_trusted"))
    # Don't test schedule/import/log as they have side effects

    # --- Section 5: Monitoring commands (v0.16.6 features) ---
    print("\n--- 5. Monitoring commands (upstream v0.16.6) ---")
    run("network.open_files", "network.open_files")
    run("throttle.max_unchoked_uploads", "throttle.max_unchoked_uploads")
    run("throttle.max_unchoked_downloads", "throttle.max_unchoked_downloads")
    run("network.total_handshakes", "network.total_handshakes")

    # --- Section 6: Status bar throttle commands (ruTorrent needs these) ---
    print("\n--- 6. Status/throttle commands for ruTorrent ---")
    run("throttle.global_up.max_rate", "throttle.global_up.max_rate")
    run("throttle.global_down.max_rate", "throttle.global_down.max_rate")

    # --- Section 7: RPC transport setters blocked for untrusted ---
    print("\n--- 7. RPC transport setters blocked for untrusted ---")
    run("network.rpc.use_xmlrpc getter (untrusted)", "network.rpc.use_xmlrpc",
        untrusted=True)
    run("network.rpc.use_jsonrpc getter (untrusted)", "network.rpc.use_jsonrpc",
        untrusted=True)
    run("network.rpc.use_xmlrpc.set blocked", "network.rpc.use_xmlrpc.set",
        EMPTY_TARGET + int_param(0),
        untrusted=True, expect_blocked=True)
    run("network.rpc.use_jsonrpc.set blocked", "network.rpc.use_jsonrpc.set",
        EMPTY_TARGET + int_param(0),
        untrusted=True, expect_blocked=True)

    # --- Summary ---
    print("\n" + "=" * 60)
    print(f"Results: {passed}/{total} passed, {failed} failed")
    print("=" * 60)

    return 0 if failed == 0 else 1


if __name__ == "__main__":
    sys.exit(main())
```

xirvik:
> This looks like it only blacklists commands at the root level, while the way rtorrent handles things you can have nested commands / expand arguments.

Thanks! Gave it another shot. Let me know

src/rpc/scgi_task.cc
```diff
@@ -287,30 +293,35 @@ SCgiTask::receive_call(const char* buffer, uint32_t length) {
...
+    break;
+  default:
+    throw torrent::internal_error("SCgiTask::receive_call(...) received bad input.");
+  }
+
+  torrent::main_thread::thread()->callback_interrupt_pollling(this, [buffer, length, result_callback, trusted, rpc_type]() {
+      rpc::RpcManager::set_trusted(trusted);
+
+      try {
+        rpc.process(rpc_type, buffer, length,
```
rakshasa: Add a `process_untrusted()` function to make it cleaner.

src/command_ui.cc
```diff
@@ -797,30 +797,30 @@ initialize_command_ui() {
...
 
-  CMD2_VAR_VALUE ("ui.throttle.global.step.small",  5);
-  CMD2_VAR_VALUE ("ui.throttle.global.step.medium", 50);
-  CMD2_VAR_VALUE ("ui.throttle.global.step.large",  500);
+  CMD2_VAR_VALUE_U("ui.throttle.global.step.small",  5);
+  CMD2_VAR_VALUE_U("ui.throttle.global.step.medium", 50);
+  CMD2_VAR_VALUE_U("ui.throttle.global.step.large",  500);
 
-  CMD2_VAR_VALUE ("ui.focus.page_size", 0);
+  CMD2_VAR_VALUE_U("ui.focus.page_size", 0);
```
rakshasa: The ui related commands aren't needed.

xirvik:
Comments addressed @rakshasa  - thanks!

src/command_events.cc
```diff
@@ -344,4 +344,24 @@ initialize_command_events() {
...
+  rpc::rpc.mark_safe("remove_untied");
+
+  rpc::rpc.mark_safe("load.normal");
+  rpc::rpc.mark_safe("load.verbose");
+  rpc::rpc.mark_safe("load.start");
+  rpc::rpc.mark_safe("load.start_verbose");
+  rpc::rpc.mark_safe("load.raw");
+  rpc::rpc.mark_safe("load.raw_verbose");
+  rpc::rpc.mark_safe("load.raw_start");
+  rpc::rpc.mark_safe("load.raw_start_verbose");
```
rakshasa: Loading torrents allow for passing commands to be executed afterwards.

This allows for bypassing the whitelist.

src/command_dynamic.cc
```diff
@@ -455,4 +455,28 @@ initialize_command_dynamic() {
   CMD2_ANY         ("strings.tracker_event",             std::bind(&torrent::option_list_strings, torrent::OPTION_TRACKER_EVENT));
   CMD2_ANY         ("strings.tracker_mode",              std::bind(&torrent::option_list_strings, torrent::OPTION_TRACKER_MODE));
   // clang-format on
+
+#ifdef HAVE_XMLRPC_TINYXML2
+  rpc::rpc.mark_safe("system.listMethods");
+#endif
+
+  rpc::rpc.mark_safe("method.use_deprecated");
+  rpc::rpc.mark_safe("method.use_deprecated.set");
```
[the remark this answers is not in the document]
xirvik: > While not currently used, it shouldn't be exposed.

Removed. Thanks!

rakshasa:
Looks good.

xirvik:
> Looks good.

Anything else needed before it can be merged?

rakshasa:
Nothing else, I'll merge it once i got some other stuff done.

rakshasa:
On reflection, this should probably have a command to set the default for SCGI to be untrusted. And add a trusted header.

Otherwise if the proxy is misconfigured and drops the header flag then the rtorrent instance would allow full access.

xirvik:
> On reflection, this should probably have a command to set the default for SCGI to be untrusted. And add a trusted header.
> 
> Otherwise if the proxy is misconfigured and drops the header flag then the rtorrent instance would allow full access.

It's ruTorrent that sets the header - no proxy involved. Its plugin httprpc (primarily; there's other variants) sets that header for untrusted commands. Basically, if the command comes from the browser and is passed as-is to rtorrent, the header is set. If it's generated server-side by rtorrent, either as a result as a browser interaction (say, start a rtorrent from the UI) or by an automated command (for example, a new torrent fetched via RSS) then no header.

Of course this requires the gateway to rtorrent to implement this stuff - other UIs are likely vulnerable.

rakshasa:
Still would be more secure for rutorrent afaict, as it would be another layer of defense in case there's a rutorrent exploit that causes the header to be dropped.

In any case it's up to you, as I'm not really that into dealing with the web UI side.

ansonlos:
@xirvik Can you please indicate what changes we need to make to our setup of rtorrent (0.16.9) and rutorrent (5.2.10-hotfix)? I notice that starting with 0.16.9, clicking on "Start", "Pause", etc buttons on a torrent no longer has any effect.

PPN-SD:
> @xirvik Can you please indicate what changes we need to make to our setup of rtorrent (0.16.9) and rutorrent (5.2.10-hotfix)? I notice that starting with 0.16.9, clicking on "Start", "Pause", etc buttons on a torrent no longer has any effect.

I agree, this new behaviour must be documented given ruttorrent is untrusted by default.

rakshasa:
If that is the case, then using the trusted/untrusted in rtorrent needs to be an option that is disabled by default.

PPN-SD:
Yes, please. I have no doubt about the benefit of trust/non-trust, but this is not documented in Ruttorent either. It seems like a rather hazardous implementation.

xirvik:
> > @xirvik Can you please indicate what changes we need to make to our setup of rtorrent (0.16.9) and rutorrent (5.2.10-hotfix)? I notice that starting with 0.16.9, clicking on "Start", "Pause", etc buttons on a torrent no longer has any effect.
> 
> I agree, this new behaviour must be documented given ruttorrent is untrusted by default.

This is not true - ruTorrent has had this implementation for ages:
https://github.com/Novik/ruTorrent/blob/e839191876b8d950dc2c6617cdfb2b726979d44e/php/xmlrpc.php#L109

However it does nothing if you aren't running a version of rTorrent that supports it, see 

https://github.com/Novik/rtorrent/blob/f9f6858065479774c12a6f16437eca07ebd39012/src/rpc/scgi_task.cc#L165

The difference however is that Novik's fork used a blacklist instead of a whitelist which @rakshasa preferred.

Anyway, to answer your question about fixing it (we saw the same problem), it's in ruTorrent, specifically in plugins/httprpc/action.php.

We'll submit a PR to ruTorrent as well, but TL;DR
sed -i 's/success(false)/success(true)/g; s/("d.open","d.start")/("d.open","d.start","d.resume")/; /case "unpause"/,/break;/ s/("d.start")/("d.resume")/' plugins/httprpc/action.php

Probably does what you need to get things working normally again.

PPN-SD:
Any documentation ? What a simply end user have to do?

xirvik:
> Any documentation ? What a simply end user have to do?

Nothing. Just update ruTorrent at the same time. This has always been transparent. 
Well, once this thing is merged there.

ansonlos:
Thanks @xirvik for your quick response on this matter. That update to httprpc plugin seems to have restored the function of the buttons. Think many others will appreciate that PR to ruTorrent. Cheers!

PPN-SD:
> Nothing. Just update ruTorrent at the same time. This has always been transparent. 
> Well, once this thing is merged there. 
> 

Tested rtorrent 0.16.9 + ruttorrent (master) + scgi socket => untrusted

It requires an option in rtorrent and a documentation.

PPN-SD:
@rakshasa Could you please revert it and publish a new release? This should be reworked.

xirvik:
> > Nothing. Just update ruTorrent at the same time. This has always been transparent.
> > Well, once this thing is merged there.
> 
> Tested rtorrent 0.16.9 + ruttorrent (master) + scgi socket => untrusted
> 
> It requires an option in rtorrent and a documentation.

I disagree; with updated rtorrent + updated ruTorrent it works same as always; except without a truck sized security hole. 

What do you want to document? Unless the header UNTRUSTED_CONNECTION is present (and it's only present if ruTorrent adds it), this is no-op for rtorrent. 

If you are using ruTorrent, then just update it as soon as this is merged there. Don't update one thing but not the other.

PPN-SD:
>except without a truck sized security hole ...

... that however makes it work!

Please don’t say that it doesn’t change anything when the comments here prove otherwise. Patch in comment can't be a fix. whitelist/blacklist: where, for what, config-file ?

I don’t understand, you’re implementing something for ruTorrent when it’s not ready there.

xirvik:
> Please don’t say that it doesn’t change anything when the comments here prove otherwise. Patch in comment can't be a fix. whitelist/blacklist: where, for what, config-file ?

It doesn't change anything from the point of view of users. 
Why do you insist on a config file? You don't configure it. There's a list of whitelisted command that ruTorrent knows to be safe: Specifically, either the ones it creates itself. And everything else (i.e. what comes from the browser) is untrusted, but if it's whitelisted (i.e. we know that they can't be abused), it doesn't matter if it's untrusted. The list of safe commands is hardcoded in both projects. There's nothing to configure. 

> 
> I don’t understand, you’re implementing something for ruTorrent when it’s not ready there.

It's not merged, but it exists. We're using it. What do you want, simultaneous merges on both projects? It will be merged this week in ruTorrent.

Anyway I sent the patch here, will send it to ruTorrent, explained what's going on, not much else to do here. If you want to revert, revert. Up to you to run a known vulnerable version of software because you can't wait a couple of days.

PPN-SD:
>  you can't wait a couple of days.

I mean _you_ can't wait. rtorrent is released, not rutorrent and now it's broken.
Please share the PR here.

xirvik:
> > you can't wait a couple of days.
> 
> I mean _you_ can't wait. rtorrent is released, not rutorrent and now it's broken. Please share the PR here.

Me? I have no rush. I'm the only that has been working on this for a while with both projects, and I'm upstreaming it both because it's important to the community, even if you think it's best to revert a security fix.

https://github.com/Novik/ruTorrent/pull/3009

Didn't take much to engage in a discussion. I'll leave it here.

rakshasa:
Since without the `UNTRUSTED_CONNECTION` header flag rtorrent works as normal, I'll consider this an rutorrent issue

PPN-SD:
>  you think it's best to revert a security fix.

Please, no (bad) shortcut. It just should have been handled in ruTorrent first, that’s all. Fixed now and hopefully with a new version of ruTorrent, thanks for this.

xirvik:
> > you think it's best to revert a security fix.
> 
> Please, no (bad) shortcut. It just should have been handled in ruTorrent first, that’s all. Fixed now and hopefully with a new version of ruTorrent, thanks for this.

I'm sorry. You have the software package relationship completely backwards. You basically don't want rtorrent making changes that could break your existing setup. That's like expecting rtorrent's dependencies not to make changes until rtorrent is ready. So no changes in libcurl or the C++ compiler.

That's not how it works.

If any piece of software makes breaking changes, which of course happens every day, it's the job of the packages that depend on that one to keep up (or stay with the previous version, that's what pinning is for).

ruTorrent can't be updated until rtorrrent changes.

ruTorrent is just one of the several clients that use rtorrent as the backend. It's not special. rtgui for example doesn't work anymore (unless patched) because rtorrent renamed commands. 

And Flood to mention another for a long time didn't work with upstream rtorrent (jesec had its own fork). 

So as a user, what can do you? Obviously, stay in the version that works for you until all the components that work together are updated. Simple as that. 

You don't go around telling the maintainers to revert anything. Basically you wanted @rakshasa to revert this patch so 0.16.9 goes back to the same thing 0.16.8 had, and then release 0.16.10 which would be the same as 0.16.8 since you removed the new stuff.

You can just stay in 0.16.8 until ruTorrent is updated.

Then you asked Novik to release a new ruTorrent version for your convenience. With that, the current develop branch, which has in-flight things that are not stable? You would then complain about those things...

Please be respectful to maintainers and contributors. If something doesn't work you just open an issue. PR comments are not for this. 

Sorry for the rant.

PPN-SD:
It’s just unpleasant to be told that "it doesn’t change anything" when that’s obviously not the case. That's my point.

> So no changes in libcurl or the C++ compiler. That's not how it works.

In the case of libs, functions are usually deprecated and then removed.

> Then you asked Novik to release a new ruTorrent version for your convenience. With that, the current develop branch, which has in-flight things that are not stable? You would then complain about those things...

Request a new version to adapt to the only software for which rutorrent is used makes sense. And this has already been done.
Publishing a new version of rutorrent doesn’t require merging everything from develop.

> Please be respectful to maintainers and contributors.

??

Issue #1712, opened 2026-03-01
Fix column widths for large torrents by allowing adjustable / configurable bandwidth and size units


d-throttle:
I would suggest auto-increasing the width of all rows/items in a column to that of the largest torrent.

![Image](https://github.com/user-attachments/assets/fec81e2f-3ffc-42e1-b220-ef2dfdf62e20)

ghost:
One solution would be to have the units change from MB -> GB when MB exceeds 999.9 MB and likewise go from GB -> TB when GB exceeds 999.9 GB. This would apply to: `Downloaded`, `Size`, and `Uploaded`. It would be nice to set base units in the config so if I always wanted GB or TB, I could choose these over MB. The default option would be AUTO where it behaves in in autoscaling fashion as I just described.

Similarly, a config option to set `Up Rate` and `Down rate` in either Mbps or MB/s and Gbps or GB/s would also be useful. AUTO behavior here too would be the default.

rakshasa:
Easiest solution is to just add 2 extra width to those columns, so if someone makes a PR i'll merge it.

Issue with using GB is that it's difficult to see that the torrent is larger than the others.

ghost:
@rakshasa here's a PR for your suggested solution to the column widths: https://github.com/rakshasa/rtorrent/pull/1720

Would you also consider a PR that allows the configuration to set the up and down speed units as I suggested earlier?

rakshasa:
Sure.

PR #1720, opened 2026-03-11
Adds configurable bandwidth units

ghost:
Adjusts the column widths for Size, Downloaded, and Uploaded to account for larger size torrents. 
See: https://github.com/rakshasa/rtorrent/issues/1712#issuecomment-4038664578

### DIFF
--- a/doc/old/rtorrent.1
+++ b/doc/old/rtorrent.1
@@ -324,7 +324,8 @@ commands are available.
 \fBupload_rate = \fIKB\fB\fR
 .TP
 \fBdownload_rate = \fIKB\fB\fR
-Set the maximum global uploand and download rates.
+Set the maximum global upload and download rates. When both are
+disabled, the throttle section is hidden from the status bar.
 .TP
 \fBmin_peers = \fIvalue\fB\fR
 .TP
--- a/doc/old/rtorrent.1
+++ b/doc/old/rtorrent.1
@@ -450,7 +450,7 @@ number of criteria, including zero, from the following:
 Change the key-bindings.
 .TP
 \fBui.torrent_list.rate_unit = \fIKB/s|MB/s|Mb/s|GB/s|Gb/s\fB\fR
-Set the rate unit for the download list. Defaults to \fBKB/s\fR\&.
+Set the rate unit for the download list and footer. Defaults to \fBKB/s\fR\&.
 .SH "FILE-SYSTEM SETTINGS"
 .PP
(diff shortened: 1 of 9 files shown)


rakshasa:
Don't change the PR, the original was something I was going to merge, this needs more review so should be a separate PR.

rakshasa:
Please re-do the original PR (as a new PR), and I'll review this one. Also make sure to write proper informative commit messages or else I'll need to squash the merge and write my own. (which removes you as author)

ghost:
Apologies, I will break the work into two different PRs.

ghost:
@rakshasa new PR: https://github.com/rakshasa/rtorrent/pull/1721 for the original change ... I made this one a draft as I continue to iterate on the solution.Also, I'm not concerned about attribution, so squash away... but I'll try to make the commit msg's a little more informative ;)

ghost:
Screenshot 2026-03-11 at 2 31 36 PM Large

This PR should be ready to test as well. However, I will re-base it once the other is merged.

As you can see from the screenshot, units reflect the setting in `.rtorrent.rc` correctly. The numerical values can be verified by using `Mb/s` and then running a `sudo iftop` command on another window which will show the speeds in Mb/s as well. Simple math to divide by 8 for Mb/s -> MB/s and similarly divide again by 1000 for GB/s.

ghost:
This PR was updated to standardize the rate calculation into a single function `get_rate_unit_info` so that it can be used not only for individual rows but for global up and down rates as well. Additionally, the precision in terms of decimal places for the global up and down rates will float based on their values to always maintain 5 characters in width. Lastly, the throttle section also gains the new bandwidth unit configuration capability. Now all bandwidth units across the entire UI are standardized via one configuration option: `ui.torrent_list.rate_unit`
  
See screenshots:
a
b

d-throttle:
@rakshasa reminder, this PR also holds the contents of [my other PR here](https://github.com/rakshasa/rtorrent/pull/1721).
2 birds 1 stone if you accept this one.
Looking forward to your feedback on these changes.

d-throttle:
I updated the PR (with the last commit) so that the entire throttle section is hidden from the status bar when both the up and down throttles are disabled. I never throttle, so I prefer not to see this section. I can imagine other seedbox users would feel the same way. @rakshasa 
a

rakshasa:
Closing this as it's not going to be merged, still will consider the simpler PR.

PR #1730, opened 2026-03-25
Add support for SCGI systemd socket activation

angularbanjo:
Add a new command, `network.scgi.open_systemd`, that binds to a file descriptor passed in via [systemd socket activation](https://www.freedesktop.org/software/systemd/man/latest/sd_listen_fds.html#). This command is mutually exclusive with the existing `network.scgi.open_*` commands, as they are already mutually exclusive with each other.

Currently, this is implemented by linking against `libsystemd`, and since this is a notable new dependency, is gated behind a default-off compile-time flag. However, the actual implementation of parsing `$LISTEN_FDS` is not especially complex, so if it is preferable, this `libsystemd` dependency can be removed in favor of a local reimplementation.

Assisted by Claude Opus 4.6, but I've reviewed the resulting code carefully and helped it follow existing rtorrent conventions. I've also tested manually in a fedora container using `systemd-socket-activate`.

### DIFF
--- a/src/command_network.cc
+++ b/src/command_network.cc
@@ -197,7 +197,7 @@ apply_scgi_systemd() {
   scgi_thread::set_scgi(scgi);
   return torrent::Object();
 #else
-  throw torrent::input_error("rtorrent was not built with systemd support.");
+  throw torrent::input_error("Systemd SCGI endpoint is not supported.");
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


angularbanjo:
Thanks for the feedback! I've pushed some updates and opened rakshasa/libtorrent#665 to add the flag read-then-write logic there instead.

smj-cc:
If `rtorrent` becomes dependent upon `libsystemd`, I will no longer be able to use it.
I'm not alone here.

For security reasons, neither `systemd` nor `dbus` are permitted on any of the systems I administer.

rakshasa:
It requires a configure flag to build in, it is by default off and thus not built in unless the user explicitly enables it.

Rtorrent will never add dependencies requirements or enable by default this.

PR #1749, opened 2026-04-23
fix off-by-one in SCgiTask::event_read

### LINKED ISSUE #1748: 0.16.10 rpc::SCgiTask::event_read crash on startup
I'm getting this reliably with 0.16.10, I think it's caused by an off-by-one in rpc::SCgiTask::event_read.

```
/usr/include/c++/14/bits/stl_vector.h:1130: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](size_type) [with _Tp = char; _Alloc = std::allocator<char>; reference = char&; size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Thread 6 "rtorrent-scgi" received signal SIGABRT, Aborted.
0x00007ffff797fb2c in pthread_kill () from /lib/x86_64-linux-gnu/libc.so.6
(gdb) bt
#0  0x00007ffff797fb2c in pthread_kill () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x00007ffff792627e in raise () from /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007ffff79098ff in abort () from /lib/x86_64-linux-gnu/libc.so.6
#3  0x00007ffff7c0290d in std::__glibcxx_assert_fail(char const*, int, char const*, char const*) () from /lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00005555556a5520 in std::vector<char, std::allocator<char> >::operator[] (this=0x555555815180, __n=<optimized out>) at /usr/include/c++/14/bits/stl_vector.h:1128
#5  std::vector<char, std::allocator<char> >::operator[] (this=0x555555815180, __n=<optimized out>) at /usr/include/c++/14/bits/stl_vector.h:1128
#6  rpc::SCgiTask::event_read (this=0x555555815120) at rpc/scgi_task.cc:98
#7  0x00007ffff7e15645 in torrent::net::Poll::process (this=0x5555557be000) at net/poll_epoll.cc:239
#8  0x00007ffff7e15872 in torrent::net::Poll::do_poll (this=<optimized out>, timeout_usec=<optimized out>) at net/poll_epoll.cc:182
#9  0x00007ffff7e29d02 in torrent::system::Thread::event_loop (this=this@entry=0x5555557ceb60) at /usr/include/c++/14/bits/unique_ptr.h:193
#10 0x00007ffff7e2a419 in torrent::system::Thread::enter_event_loop (thread=0x5555557ceb60) at system/thread.cc:168
#11 0x00007ffff797daa4 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#12 0x00007ffff7a0ac6c in ?? () from /lib/x86_64-linux-gnu/libc.so.6
```

SCgiTask::open() sizes m_buffer as default_buffer_size (8192) plus 1 to account for null termination.

SCgiTask::read() uses size() in the recv() call which only sees a 8193 byte buffer.

If recv() fills the buffer, m_position points past the end, and m_buffer[m_position] = '\0' goes boom.

fffe:
Reads one byte less if m_content_length is 0. Fixes #1748.

### DIFF
--- a/src/rpc/scgi_task.cc
+++ b/src/rpc/scgi_task.cc
@@ -80,7 +80,12 @@ SCgiTask::close() {
 
 void
 SCgiTask::event_read() {
-  int bytes = ::recv(m_fileDesc, m_buffer.data() + m_position, m_buffer.size() - m_position - (m_content_length == 0), 0);
+  int read_length = m_buffer.size() - m_position;
+  
+  if (m_content_length == 0)
+    read_length--;
+  
+  int bytes = ::recv(m_fileDesc, m_buffer.data() + m_position, read_length, 0);
 
   if (bytes <= 0) {
     if (bytes == 0 || !(errno == EAGAIN || errno == EINTR))
--- a/src/rpc/scgi_task.cc
+++ b/src/rpc/scgi_task.cc
@@ -84,6 +84,9 @@ SCgiTask::event_read() {
   
   if (m_content_length == 0)
     read_length--;
+
+  if (read_length <= 0)
+    throw torrent::internal_error("SCgiTask::event_read() no space in buffer for event_read.");
   
   int bytes = ::recv(m_fileDesc, m_buffer.data() + m_position, read_length, 0);
 
--- a/src/rpc/scgi_task.cc
+++ b/src/rpc/scgi_task.cc
@@ -80,7 +80,7 @@ SCgiTask::close() {
 
 void
 SCgiTask::event_read() {
-  int bytes = ::recv(m_fileDesc, m_buffer.data() + m_position, m_buffer.size() - m_position, 0);
+  int bytes = ::recv(m_fileDesc, m_buffer.data() + m_position, m_buffer.size() - m_position - (m_content_length == 0), 0);
 
   if (bytes <= 0) {
     if (bytes == 0 || !(errno == EAGAIN || errno == EINTR))


fffe:
I think https://github.com/fffe/rtorrent/blob/master/src/rpc/scgi_task.cc#L43-L46 is racy too.

The resize should come before adding it to the event loop, right?

If I'm reading that right, I can add it to this patch.

rakshasa:
It's not a race condition, the poll is in `this_thread` (scgi_thread) and the only reason we're locking the result mutex is to ensure there's a memory barrier as the data is handed over to main_thread.

Strictly speaking we don't need to a memory barrier there, only right before main_thread callback is pushed.

rakshasa:
```
int read_length = m_buffer.size() - m_position;

if (m_content_length == 0)
  read_length--;

int bytes = ::recv(m_fileDesc, m_buffer.data() + m_position, read_length, 0);
```

This is more readable.

fffe:
Sure, updated.

rakshasa:
We probably need to throw internal_error if `read_length == 0` before the check.

fffe:
Shouldn't it be read_length <= 0 after the check?

You want to catch read_length = 0 and read_length == 1 && m_content_length == 0

rakshasa:
For future reference, commit messages should start with capital letter and end in punctuation, and be descriptive.

Use force-push when updating PRs.

PR #1756, opened 2026-05-04
Move d.save_full_session to run after user-registered inserted_new handlers

xirvik:
Since 0.16.6 moved session saving to a separate thread, `save_full_download()` synchronously snapshots the bencode when called and queues prebuilt streamsfor async write.

If user-registered handlers (e.g. `d.set_custom=addtime,$execute_capture=...`) sort lexicographically AFTER `1_prepare`, they modify the bencode AFTER the snapshot has been taken. The first on-disk `.rtorrent` then lacks those custom fields. The next periodic resume save catches up — but if rtorrent restarts before that, the data is lost.

This manifests in ruTorrent as blank Finished/SeedingTime/AddTime columns for some torrents (Novik/ruTorrent#3050).

## Test

Verified on a downstream patched 0.16.11 build with ruTorrent: addtime present in `.rtorrent` immediately after add, mtime matches add time.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -209,7 +209,14 @@ main(int argc, char** argv) {
        "method.insert = event.download.hash_queued,multi|rlookup|static\n"
 
        "method.set_key = event.download.inserted,         1_send_scrape, ((d.tracker.send_scrape,30))\n"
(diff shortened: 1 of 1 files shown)


rakshasa:
```
       "method.set_key = event.download.erased, ~_delete_tied, d.delete_tied=\n"
```

The `~` is the last character, so `~_save_full` should be used imo.

Issue #1757, opened 2026-05-04
directory.watch.added fails for normal macOS Finder copies to SMB watch directories


tonimelisma:
## Summary

`directory.watch.added` does not work reliably for an rTorrent watch directory exposed as an SMB share and used from macOS Finder.

Here's how to reproduce:

1. Export an rTorrent watch directory over Samba.
2. From macOS Finder, copy a `.torrent` file into that SMB share.
3. Finder/Samba creates the final `.torrent` filename before the file contents are available.
4. rTorrent loads immediately on the early `IN_CREATE` event, while the file is empty.
5. The file copy completes shortly afterwards, but rTorrent does not retry the completed file.

On my system, each `.torrent` copied this way fails to load.

## Reproducer

I reproduced this with rTorrent/libtorrent `0.15.7` on Linux. Current `master` appears to have the same relevant code path.

The watch directory is a local Linux directory exported by Samba and mounted from macOS. rTorrent is configured with:

```ini
directory.watch.added = /home/toni/Torrents, load.start_verbose
```

Copying a `.torrent` file from macOS Finder into that SMB share logs:

```text
Reading torrent file failed: "/home/toni/Torrents/final-name.torrent"
```

The file is valid after the copy completes.

## Host-side evidence

`inotifywait` on the Linux host showed this sequence for the normal Finder copy:

```text
CREATE final-name.torrent
ATTRIB final-name.torrent
CLOSE_WRITE,CLOSE final-name.torrent
MODIFY final-name.torrent
CLOSE_WRITE,CLOSE final-name.torrent
```

`strace` attached to rTorrent showed that rTorrent reacted to the early `CREATE` event:

```text
read(inotify_fd, "...final-name.torrent...", 2048) = 128
newfstatat(..., "final-name.torrent", {st_size=0, ...}, 0) = 0
openat(..., "final-name.torrent", O_RDONLY) = fd
read(fd, "", 8191) = 0
close(fd) = 0
```

Later close-write events saw the final file size, but rTorrent did not reopen/load the torrent.

## Current code path

From current `master`:

- libtorrent `directory_events::notify_on()` maps `flag_on_added` to `IN_CREATE | IN_MOVED_TO`, and `flag_on_updated` to `IN_CLOSE_WRITE`.
- libtorrent `directory_events::event_read()` filters for `.torrent` filenames, but does not pass the inotify mask to the callback.
[...]

PR #1758, opened 2026-05-04
Add directory.watch.ready for safe torrent watch folders

tonimelisma:
## Summary

Add `directory.watch.ready`, an opt-in watch-directory command that loads `.torrent` files when they are plausibly ready to be read, including normal copies into network shares.

Example:

```ini
directory.watch.ready = /home/user/Torrents/watch, load.start_verbose
```

## Behavior

`directory.watch.ready` uses libtorrent's ready directory event support, intended to trigger on candidate-ready events:

- completed files moved/renamed into the watch directory;
- final-name files closed after writing.

rTorrent then queues those callbacks briefly until the target path is:

- present;
- a regular file;
- non-empty;
- stable across a quiet interval.

This deliberately avoids loading on initial create events, because normal file-copy workflows can create the final filename before the contents are available. This is reproducible with macOS Finder copying a `.torrent` file into an SMB share watched by rTorrent.

Readiness is best-effort rather than a kernel guarantee, but this covers the two common watch-folder workflows: temp-name-then-rename and final-name-write-then-close.

## Compatibility

- Existing `directory.watch.added` behavior is unchanged.
- `directory.watch.ready` is opt-in.
- Mixed `directory.watch.added` / `directory.watch.ready` registrations for the same canonical watch directory are rejected by libtorrent's directory event registration.
- Parent/child and sibling watch directories are allowed, because inotify directory watches are not recursive.

## Implementation Notes

- rTorrent checks at configure time that libtorrent provides `torrent::directory_events::flag_on_ready`.
- Ready-watch callbacks are queued before dispatching the configured load command.
- Missing, zero-length, or non-regular stale paths are dropped instead of waking forever.
- Queued ready-watch work is cleared/disabled during shutdown.
- The local file-status cache includes size as well as mtime so same-second replacements with different contents are not treated as unchanged.

## Dependency Ordering

The corresponding libtorrent ready-watch support landed in rakshasa/libtorrent#725.

This rTorrent PR still keeps a compile-time `directory_events::flag_on_ready` probe because the released libtorrent `v0.16.12` tag does not expose that API, while current libtorrent master still reports version `0.16.12`. A version check alone is therefore ambiguous until libtorrent gets a later release/version bump containing the ready-watch API.

## Rationale

`directory.watch.added` can fail for normal macOS Finder copies to SMB watch directories:

1. Finder/Samba creates the final `.torrent` path.
2. rTorrent loads on `IN_CREATE` while the file is empty.
3. The copy completes.
4. rTorrent does not retry the completed file.

`directory.watch.ready` gives users a single safe command for normal watch-folder use without requiring them to understand inotify event names.

Addresses: #1757.

## Verification

- `./configure` passes against libtorrent with ready-watch support.
- Fake released libtorrent `v0.16.12` headers fail the `flag_on_ready` probe as intended.
- `git diff --check` passes.
- Clean `bear -- make` passes.
- `clang-tidy-diff` has no changed-line diagnostics.
- `make check` passes (`2/2`).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -52,6 +52,19 @@ PKG_CHECK_MODULES([ZLIB], [zlib])
 PKG_CHECK_MODULES([DEPENDENCIES], [libtorrent >= 0.16.12])
 
 AC_LANG_PUSH(C++)
+rtorrent_save_CXXFLAGS="$CXXFLAGS"
+CXXFLAGS="$CXXFLAGS $DEPENDENCIES_CFLAGS"
+AC_MSG_CHECKING([for libtorrent directory_events ready watch support])
+AC_COMPILE_IFELSE(
+  [AC_LANG_PROGRAM([[#include <torrent/utils/directory_events.h>]],
+                   [[return torrent::directory_events::flag_on_ready == 0;]])],
+  [AC_MSG_RESULT([yes])],
+  [AC_MSG_RESULT([no])
+   AC_MSG_ERROR([requires libtorrent with directory_events::flag_on_ready])])
+CXXFLAGS="$rtorrent_save_CXXFLAGS"
+AC_LANG_POP(C++)
+
+AC_LANG_PUSH(C++)
 TORRENT_WITH_XMLRPC_C
 AC_LANG_POP(C++)
 
--- a/src/command_events.cc
+++ b/src/command_events.cc
@@ -2,6 +2,8 @@
 
 #include <functional>
 #include <cstdio>
+#include <string>
+#include <vector>
 #include <torrent/rate.h>
 #include <torrent/hash_string.h>
 #include <torrent/utils/log.h>
@@ -19,6 +21,7 @@
 #include "rpc/command_scheduler.h"
 #include "rpc/parse.h"
 #include "rpc/parse_commands.h"
+#include "utils/watch_ready_queue.h"
 
 torrent::Object
 apply_on_ratio(const torrent::Object& rawArgs) {
@@ -291,20 +294,37 @@ call_watch_command(const std::string& command, const std::string& path) {
 }
 
 torrent::Object
-directory_watch_added(const torrent::Object::list_type& args) {
+directory_watch(const torrent::Object::list_type& args, int flags) {
   if (args.size() != 2)
     throw torrent::input_error("Too few arguments.");
 
   auto& path = args.front().as_string();
   auto& command = args.back().as_string();
+  std::string expanded_path = expand_path(path);
 
   if (!control->directory_events()->open())
     throw torrent::input_error("Could not open inotify:" + std::string(std::strerror(errno)));
 
-  control->directory_events()->notify_on(path.c_str(),
-                                         torrent::directory_events::flag_on_added | torrent::directory_events::flag_on_updated,
-                                         std::bind(&call_watch_command, command, std::placeholders::_1));
+  torrent::watch_descriptor::slot_string slot =
+    flags == torrent::directory_events::flag_on_ready ?
+    torrent::watch_descriptor::slot_string([command](const auto& arg) { control->watch_ready_queue()->push(command, arg); }) :
+    torrent::watch_descriptor::slot_string(std::bind(&call_watch_command, command, std::placeholders::_1));
+
+  control->directory_events()->notify_on(expanded_path.c_str(), flags, slot);
   return torrent::Object();
+}
+
+torrent::Object
+directory_watch_added(const torrent::Object::list_type& args) {
+  return directory_watch(args,
+                         torrent::directory_events::flag_on_added |
+                         torrent::directory_events::flag_on_updated);
+}
+
+torrent::Object
+directory_watch_ready(const torrent::Object::list_type& args) {
+  return directory_watch(args,
+                         torrent::directory_events::flag_on_ready);
 }
 
 void
@@ -344,6 +364,7 @@ initialize_command_events() {
   CMD2_ANY_LIST    ("d.multicall.filtered", std::bind(&d_multicall_filtered, std::placeholders::_2));
 
   CMD2_ANY_LIST    ("directory.watch.added", std::bind(&directory_watch_added, std::placeholders::_2));
+  CMD2_ANY_LIST    ("directory.watch.ready", std::bind(&directory_watch_ready, std::placeholders::_2));
 
   rpc::rpc.mark_safe("start_tied");
   rpc::rpc.mark_safe("stop_untied");
--- a/src/utils/watch_ready_queue.cc
+++ b/src/utils/watch_ready_queue.cc
@@ -1,0 +1,158 @@
+#include "config.h"
+
+#include "utils/watch_ready_queue.h"
+
+#include <algorithm>
+#include <torrent/utils/file_stat.h>
+#include <torrent/utils/log.h>
+
+#include "globals.h"
+#include "rpc/rpc_manager.h"
+
+namespace utils {
+
+namespace {
+
+auto compare_next_time = [](const auto* lhs, const auto* rhs) {
+  return lhs->next_time > rhs->next_time;
+};
+
+}
+
+WatchReadyQueue::WatchReadyQueue() {
+  m_task_process.slot() = [this]() { process(); };
+}
+
+WatchReadyQueue::~WatchReadyQueue() {
+  torrent::this_thread::scheduler()->erase(&m_task_process);
+}
+
+void
+WatchReadyQueue::push(const std::string& command, const std::string& path) {
+  if (!m_active)
+    return;
+
+  auto result = m_entries.emplace(path, Entry());
+  auto& entry = result.first->second;
+
+  if (result.second) {
+    entry.path = path;
+    entry.first_seen = torrent::this_thread::cached_time();
+  }
+
+  entry.command = command;
+  entry.last_changed = torrent::this_thread::cached_time();
+  update_status(&entry);
+
+  if (result.second) {
+    update_next_time(&entry);
+    push_entry(&entry);
+  } else {
+    update_entry(&entry);
+  }
+
+  schedule();
+}
+
+void
+WatchReadyQueue::shutdown() {
+  m_active = false;
+  m_entries.clear();
+  m_entry_queue.clear();
+  torrent::this_thread::scheduler()->erase(&m_task_process);
+}
+
+void
+WatchReadyQueue::push_entry(Entry* entry) {
+  m_entry_queue.push_back(entry);
+  std::push_heap(m_entry_queue.begin(), m_entry_queue.end(), compare_next_time);
+}
+
+void
+WatchReadyQueue::update_entry(Entry* entry) {
+  update_next_time(entry);
+  std::make_heap(m_entry_queue.begin(), m_entry_queue.end(), compare_next_time);
+}
+
+void
+WatchReadyQueue::update_next_time(Entry* entry) {
+  if (entry->regular && entry->size > 0) {
+    entry->next_time = entry->last_changed + quiet_time;
+    return;
+  }
+
+  entry->next_time = std::min(torrent::this_thread::cached_time() + retry_time,
+                              entry->first_seen + stale_time);
+}
+
+void
+WatchReadyQueue::update_status(Entry* entry) {
+  torrent::utils::FileStat fs;
+  bool regular = false;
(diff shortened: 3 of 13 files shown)


rakshasa:
That's a lot of code, some of it generalizing the code unnecessarily and the collision check should be done in `control->directory_events()`.

tonimelisma:
Thanks, noted. I'm still working on this.

tonimelisma:
Updated this per your feedback.

The rTorrent-side watch-root tracking/generalized collision code has been removed. Mixed-mode collision handling now lives in `control->directory_events()` / libtorrent, where the watch descriptors are registered.

The rTorrent side now does three things:

- adds the opt-in `directory.watch.ready` command
- checks at configure time that libtorrent provides `directory_events::flag_on_ready`
- queues ready-watch callbacks briefly until the target file is non-empty, regular, and stable before calling the configured load command

The queue is intentional: low-level `IN_MOVED_TO | IN_CLOSE_WRITE` events are enough to avoid loading on `IN_CREATE`, but they are not by themselves a perfect file-completion signal. When I tested this on the latest macOS, moving a torrent file to a Samba share, it produced multiple close-write events. Hence we need rTorrent to wait for a stable file state before dispatching the load command and drop stale unready paths instead of waking forever. Without this behavior, I was able to get torrents loaded but the client would still produce errors because it triggered too soon on the 0 byte file. I believe this is standard for Samba shares.

This rTorrent PR is intentionally stacked on rakshasa/libtorrent#697: libtorrent needs to land first, then this PR can build against upstream libtorrent. CI won't pass as-is.

Existing `directory.watch.added` behavior is unchanged.

Verification: `make check` passes locally.

src/utils/watch_ready_queue.cc
```diff
@@ -0,0 +1,177 @@
...
+  auto now = torrent::this_thread::cached_time();
+  auto earliest = next_time(m_entries.begin()->second, now);
+
+  for (auto itr = std::next(m_entries.begin()); itr != m_entries.end(); ++itr)
+    earliest = std::min(earliest, next_time(itr->second, now));
+
+  if (m_task_process.is_scheduled())
+    torrent::this_thread::scheduler()->update_wait_until(&m_task_process, earliest);
+  else
+    torrent::this_thread::scheduler()->wait_until(&m_task_process, earliest);
```
rakshasa: Use only `update_wait_until` as it does all that checking already.

src/utils/watch_ready_queue.cc
```diff
@@ -0,0 +1,177 @@
...
+    entry->m_last_changed = now;
+  }
+}
+
+void
+WatchReadyQueue::process() {
+  if (m_disabled || (control != nullptr && control->is_shutdown_started())) {
+    clear();
+    return;
+  }
```
rakshasa: This is never reached, or should never be reached.

src/utils/watch_ready_queue.cc
```diff
@@ -0,0 +1,177 @@
...
+
+constexpr auto watch_ready_quiet_time = std::chrono::milliseconds(500);
+constexpr auto watch_ready_retry_time = std::chrono::milliseconds(250);
+constexpr auto watch_ready_stale_time = std::chrono::seconds(10);
+
+void
+log_dropped_unready_file(const std::string& path) {
+  if (control != nullptr && control->core() != nullptr)
+    control->core()->push_log_std("Dropped unready watch file after timeout: \"" + path + "\"");
+}
```
rakshasa, line 20: Use `torrent::LOG_NOTICE` for the same effect, and no need for the if checks.

src/utils/watch_ready_queue.cc
```diff
@@ -0,0 +1,177 @@
...
+
+} // namespace
+
+WatchReadyQueue::WatchReadyQueue() {
+  m_task_process.slot() = std::bind(&WatchReadyQueue::process, this);
```
rakshasa: Use `[this]() { process(); }`

src/utils/watch_ready_queue.cc
```diff
@@ -0,0 +1,177 @@
...
+
+void
+WatchReadyQueue::clear() {
+  m_entries.clear();
+  torrent::this_thread::scheduler()->erase(&m_task_process);
+}
+
+void
+WatchReadyQueue::disable() {
+  m_disabled = true;
```
rakshasa: Rename to `m_active` to be consistent, and call it `shutdown()` and remove `clear()` as it is unneeded.

src/utils/watch_ready_queue.cc
```diff
@@ -0,0 +1,177 @@
...
+
+  for (auto itr = m_entries.begin(); itr != m_entries.end(); ) {
+    Entry& entry = itr->second;
+    update_status(&entry, now);
+
+    bool nonempty_regular = entry.m_status.m_exists &&
+                            entry.m_status.m_regular &&
+                            entry.m_status.m_size > 0;
+    bool stable = now - entry.m_last_changed >= watch_ready_quiet_time;
+    bool stale = now - entry.m_first_seen >= watch_ready_stale_time;
```
rakshasa: Rename to `stale_entry` and `unchanged_entry` to be more readable.

src/utils/watch_ready_queue.cc
```diff
@@ -0,0 +1,177 @@
...
+WatchReadyQueue::~WatchReadyQueue() {
+  torrent::this_thread::scheduler()->erase(&m_task_process);
+}
+
+void
+WatchReadyQueue::push(const std::string& command, const std::string& path) {
+  if (m_disabled)
+    return;
+
+  auto now = torrent::this_thread::cached_time();
```
rakshasa: Cached_time() is a simple function call that returns a value, don't bother using temporary variables for it.

Same for `next_time()`, just call it in that function instead of passing it as an arg.

src/utils/watch_ready_queue.cc
```diff
@@ -0,0 +1,177 @@
...
+#include "globals.h"
+#include "rpc/rpc_manager.h"
+
+namespace utils {
+
+namespace {
+
+constexpr auto watch_ready_quiet_time = std::chrono::milliseconds(500);
+constexpr auto watch_ready_retry_time = std::chrono::milliseconds(250);
+constexpr auto watch_ready_stale_time = std::chrono::seconds(10);
```
rakshasa: These should be in the class declaration.

src/utils/watch_ready_queue.h
```diff
@@ -0,0 +1,62 @@
...
+  };
+
+  struct Entry {
+    std::string m_command;
+    std::string m_path;
+    FileStatus  m_status;
+    bool        m_has_status{};
+    time_type   m_first_seen{};
+    time_type   m_last_changed{};
+  };
```
rakshasa, line 43:
```
struct Entry {
   bool is_nonempty_regular() const;
   ....
}

  ...
};

inline bool is_nonempty_regular() const { return ...}
```

src/command_events.cc
```diff
@@ -290,23 +293,45 @@ call_watch_command(const std::string& command, const std::string& path) {
...
 
   if (!control->directory_events()->open())
     throw torrent::input_error("Could not open inotify:" + std::string(std::strerror(errno)));
 
-  control->directory_events()->notify_on(path.c_str(),
-                                         torrent::directory_events::flag_on_added | torrent::directory_events::flag_on_updated,
-                                         std::bind(&call_watch_command, command, std::placeholders::_1));
+  torrent::directory_events::slot_string slot =
+    flags == torrent::directory_events::flag_on_ready ?
+    torrent::directory_events::slot_string(std::bind(&queue_ready_watch_command, command, std::placeholders::_1)) :
```
rakshasa: Use `[command](auto& arg) { control->watch_ready_queue()->push(command, arg); }` to avoid unnecessary function definitions.

tonimelisma:
Thanks for the detailed review. I've addressed the inline comments.

Summary of changes:

- moved the ready-watch timing constants into `WatchReadyQueue`
- added `Entry::is_nonempty_regular()`
- simplified the ready-watch callback registration
- switched stale-entry logging to `LOG_NOTICE`
- renamed the queue shutdown state and shutdown method
- simplified scheduler updates through `update_wait_until()`
- cleaned up the ready-queue tests to match the updated API

Verification:

- `git diff --check` passes
- `make check` passes locally (`2/2` test binaries passing)

Reminder: this rTorrent PR depends on the related libtorrent PR for the new `directory_events` ready-watch support: https://github.com/rakshasa/libtorrent/pull/697

src/utils/watch_ready_queue.cc
```diff
@@ -0,0 +1,140 @@
...
+WatchReadyQueue::schedule() {
+  if (m_entries.empty()) {
+    torrent::this_thread::scheduler()->erase(&m_task_process);
+    return;
+  }
+
+  auto earliest = next_time(m_entries.begin()->second);
+
+  for (auto itr = std::next(m_entries.begin()); itr != m_entries.end(); ++itr)
+    earliest = std::min(earliest, next_time(itr->second));
```
rakshasa: This is inefficient, you should keep the vector sorted at all times.

However you're using `torrent::this_thread::cached_time() + retry_time` below, which changes value each time `schedule()` is called. (which is not desirable)

Make the timeouts well-defined and then use e.g. STL heap to keep it sorted. (see torrent/utils/scheduler.h)

tonimelisma:
Updated this to address the latest review comment about timeout ordering.

The ready-watch queue now keeps a heap of pending entries keyed by each entry’s `m_next_time`, so `schedule()` no longer scans the whole map or recomputes retry deadlines from the current time. Retry/quiet/stale deadlines are stored per entry and remain stable unless that entry’s observed file state changes.

I also added a regression test covering the specific case where an unrelated later event must not delay an earlier missing-file retry.

Verification:

- `git diff --check upstream/master..HEAD` passes
- `git diff --check origin/directory-watch-ready..HEAD` passes
- clean `bear -- make` passes
- `clang-tidy-diff` over the full PR diff produces no fixes/diagnostics on changed lines
- `make check` passes locally (`2/2` test binaries passing)

rakshasa:
Check `tonimelisma-directory-watch-ready`, not really ready for merging.

tonimelisma:
Thanks. I checked `tonimelisma-directory-watch-ready` and updated this PR from current `master`, using that branch as style guidance rather than copying it wholesale. The PR branch is now a single clean commit on top of `master`.

The main changes are:

- simplified `WatchReadyQueue` internals and moved private status/helper logic into `watch_ready_queue.cc`;
- kept retry/quiet/stale deadlines heap-ordered with per-entry `next_time`;
- folded the status fields into `Entry`;
- restored the configure-time `directory_events::flag_on_ready` probe. libtorrent ready-watch support is now upstream via rakshasa/libtorrent#725, but the released `v0.16.12` tag does not expose `flag_on_ready` while libtorrent master still reports `0.16.12`, so a version check alone is ambiguous.

One behavior I intentionally kept different: retry processing re-stats the same queued path. I did not add the cached-status early return because it can miss a zero-length file becoming non-empty without a second watch event. I added `test_retry_restats_zero_length_file_without_new_event` to cover that regression.

Verification:

- `./configure` passes against libtorrent with ready-watch support;
- fake released libtorrent `v0.16.12` headers fail the `flag_on_ready` probe as intended;
- `git diff --check` passes;
- clean `bear -- make` passes;
- `clang-tidy-diff` has no changed-line diagnostics;
- `make check` passes (`2/2`).

configure.ac
```diff
@@ -51,6 +51,19 @@ PKG_CHECK_MODULES([CPPUNIT], [cppunit],, [no_cppunit="yes"])
 PKG_CHECK_MODULES([ZLIB], [zlib])
 PKG_CHECK_MODULES([DEPENDENCIES], [libtorrent >= 0.16.12])
 
+AC_LANG_PUSH(C++)
+rtorrent_save_CXXFLAGS="$CXXFLAGS"
+CXXFLAGS="$CXXFLAGS $DEPENDENCIES_CFLAGS"
+AC_MSG_CHECKING([for libtorrent directory_events ready watch support])
+AC_COMPILE_IFELSE(
+  [AC_LANG_PROGRAM([[#include <torrent/utils/directory_events.h>]],
+                   [[return torrent::directory_events::flag_on_ready == 0;]])],
```
rakshasa, line 60: This check should not be here, we always synchronize libtorrent and rtorrent versions.

rakshasa:
Merging with #1783.

PR #1764, opened 2026-05-09
fix SCGI event_read: return on partial header read instead of closing connection

trim21:
## Summary
- Fixes a bug where rtorrent closes SCGI TCP connections if the header length integer (e.g. `"72"`) arrives without the `':'` separator in the same TCP segment
- The connection is wrongly dropped with `event_read_failed` — this causes empty responses on consecutive RPC calls

## Root Cause
In `SCgiTask::event_read()`, `std::from_chars` parses the header length (e.g. `"72"`) successfully. When only `"72"` has been received (the `':'` is in a later TCP segment), `from_chars` consumes all buffered bytes and sets `current = m_buffer.data() + m_position`.

The code then reads `*current` — but this is the null terminator manually written at `m_buffer[m_position]`, not actual received data. The check `*current != ':'` succeeds, and the code jumps to `event_read_failed`, closing the connection before the `':'` even arrives.

## Fix
Added a simple check: if `from_chars` consumed the entire buffer (`current == m_buffer.data() + m_position`), return and wait for more data via the next `event_read` callback.

```diff
     if (current == m_buffer.data())
       return;
 
+    if (current == m_buffer.data() + m_position)
+      return;
+
     if (*current != ':' || header_size < 17 || header_size > max_header_size)
       goto event_read_failed;
```

## Reproduction
Run the script below (zero dependencies, stdlib only) against a running rtorrent SCGI port:

<details>
<summary><code>repro-scgi-partial-read.py</code></summary>

```python
#!/usr/bin/env python3
"""Reproduce SCGI partial-read bug in rtorrent v0.16.11.

Sends a valid SCGI request to an rtorrent SCGI TCP port in deliberately
small TCP segments to trigger the bug where the server closes the
connection if the header length integer (e.g. "72") arrives without the
trailing ':' separator in the same segment.

Usage:
    python3 repro-scgi-partial-read.py host:port
    python3 repro-scgi-partial-read.py 127.0.0.1:5101
"""

import socket
import sys
import time

SCGI_REQUEST = (
    b"CONTENT_LENGTH\x00154\x00"
    b"SCGI\x001\x00"
)

XML_BODY = (
    b'<?xml version="1.0" encoding="UTF-8"?>'
    b"<methodCall>"
    b"<methodName>system.listMethods</methodName>"
    b"<params>"
    b"</params>"
    b"</methodCall>"
)

HEADER_LENGTH = len(SCGI_REQUEST)
HEADER_LEN_STR = str(HEADER_LENGTH).encode()


def send_in_chunks(host: str, port: int, chunks: list[bytes]) -> bytes:
    with socket.create_connection((host, port), timeout=5) as sock:
        for data in chunks:
            sock.sendall(data)
            time.sleep(0.05)  # give server time to process each segment

        return sock.recv(4096)


def test_normal(host: str, port: int) -> bool:
    """Normal send: all in one chunk (works)."""
    full_request = HEADER_LEN_STR + b":" + SCGI_REQUEST + b"," + XML_BODY
    response = send_in_chunks(host, port, [full_request])
    ok = len(response) > 0
    print(f"  Normal send: {'OK' if ok else 'FAIL'} (got {len(response)} bytes)")
    if not ok:
        print(f"    WARNING: Even normal send failed - server may not be running")
    return ok


def test_split_before_colon(host: str, port: int) -> bool:
    """Split header length from ':' (triggers the bug)."""
    chunks = [
        HEADER_LEN_STR,          # "72"
        b":" + SCGI_REQUEST + b"," + XML_BODY,
    ]
    response = send_in_chunks(host, port, chunks)
    ok = len(response) > 0
    print(f"  Split before ':': {'OK' if ok else 'FAIL'} (got {len(response)} bytes)")
    return ok


def test_split_after_colon(host: str, port: int) -> bool:
    """Split after ':' (should work)."""
    chunks = [
        HEADER_LEN_STR + b":",   # "72:"
        SCGI_REQUEST + b"," + XML_BODY,
    ]
    response = send_in_chunks(host, port, chunks)
    ok = len(response) > 0
    print(f"  Split after ':': {'OK' if ok else 'FAIL'} (got {len(response)} bytes)")
    return ok


def main() -> None:
    if len(sys.argv) != 2:
        print(__doc__)
        sys.exit(1)

    addr = sys.argv[1]
    if ":" in addr:
        host, port_str = addr.rsplit(":", 1)
        port = int(port_str)
    else:
        host, port = addr, 80

    print(f"=== Reproducing SCGI partial-read bug against {host}:{port} ===\n")

    normal_ok = test_normal(host, port)
    if not normal_ok:
        print("\nServer doesn't seem to be running or reachable. Aborting.")
        sys.exit(1)

    before_ok = test_split_before_colon(host, port)
    after_ok = test_split_after_colon(host, port)

    print()
    if not before_ok:
        print("RESULT: BUG CONFIRMED - server closes connection when ':' is in a separate TCP segment.")
    else:
        print("RESULT: Bug not reproduced - server handles split correctly.")
    print(f"  normal={normal_ok} split_before_colon={before_ok} split_after_colon={after_ok}")


if __name__ == "__main__":
    main()
```

</details>

### Without fix
```
$ python3 repro-scgi-partial-read.py 127.0.0.1:5101
  Normal send: OK (got 4012 bytes)
  Split before ':': FAIL (got 0 bytes)     <- BUG
  Split after ':': OK (got 4012 bytes)
```

### With fix
```
$ python3 repro-scgi-partial-read.py 127.0.0.1:5101
  Normal send: OK (got 4012 bytes)
  Split before ':': OK (got 4012 bytes)    <- FIXED
  Split after ':': OK (got 4012 bytes)
```

### DIFF
--- a/src/rpc/scgi_task.cc
+++ b/src/rpc/scgi_task.cc
@@ -116,6 +116,9 @@ SCgiTask::event_read() {
     if (current == m_buffer.data())
       return;
 
+    if (current == m_buffer.data() + m_position)
+      return;
+
     if (*current != ':' || header_size < 17 || header_size > max_header_size)
       goto event_read_failed;
 


trim21:
and we should return a valid response to show the errors when it failed to parse request, it's too hard to debug this

rakshasa:
Merged in https://github.com/rakshasa/rtorrent/pull/1765 due to merge conflicts.

PR #1769, opened 2026-05-12
Log tracker failures as warnings

### LINKED ISSUE #1048: LOG: Tracker log levels are buggered, only info & debug shows anything but that is too verbose for practical use
Turns out the log levels are buggered as well.
Even on latest nothing below info shows anything, and way too many things are on the info level.

Things like failed scrapes should be on the less verbose levels too imho

JustABiologist:
## Summary

Fixes #1048 by logging tracker failure messages at warning level in addition to the existing UI status message.

This makes tracker failures visible through less-verbose log outputs such as `warn`, without changing the existing tracker event/request logging behavior.

## Verification

- `git diff --check`
- Built official `libtorrent-0.16.11` release tarball into a local prefix
- Configured official `rtorrent-0.16.11` release tarball against that local prefix
- `make core/download.o`

Note: a full `make -j2` on the release tarball stopped in unrelated `command_ui.cc` template code before reaching this changed file.

### DIFF
--- a/src/core/download.cc
+++ b/src/core/download.cc
@@ -10,6 +10,7 @@
 #include <torrent/data/file.h>
 #include <torrent/data/file_list.h>
 #include <torrent/utils/file_stat.h>
+#include <torrent/utils/log.h>
 
 #include "rpc/parse_commands.h"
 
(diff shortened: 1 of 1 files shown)


rakshasa:
I'm moving everything away from those logging categories (warn, info, etc).

If you want a special tracker failure message log, create a LOG_TRACKER_FAILURE_MESSAGES category and print it out in libtorrent's tracker code. Make it formatted suitable for output to UI.

JustABiologist:
Fair sounds like the best way forward in general. Best of luck!

PR #1773, opened 2026-05-14
Honor trackers.use_udp for loaded torrents

### LINKED ISSUE #1043: trackers.use_udp.set = no seems to ENABLE udp trackers
Even if you specify no it seems to enable udp trackers.
Only commenting this line out will disable udp trackers, and result in "cannot connect to udp tracker" messages.

Version: 0.9.8 latest git / 0.13.8 udns branch. xmlrpc-c rev 3116
Distro Debian 10 with all latest patches

iamdinhthuan:
Fixes #1043.

`trackers.use_udp.set = no` already changed the global setting, but loaded torrents could keep UDP trackers enabled. This makes the setter immediately disable UDP trackers on currently loaded downloads, and applies the startup/session-load check after resume tracker settings are restored so saved tracker state cannot re-enable them.

I kept `trackers.use_udp.set = yes` as a setting change only, so it does not unexpectedly re-enable trackers a user may have disabled manually.

Validation:
- `make -j2`
- `make -C test rtorrent_Test_Src`
- `./test/rtorrent_Test_Src`
- `make check`
- `git diff --check`

### DIFF
--- a/src/command_tracker.cc
+++ b/src/command_tracker.cc
@@ -63,6 +63,18 @@ apply_dht_add_node(const std::string& arg) {
 }
 
 torrent::Object
+apply_trackers_use_udp(int64_t arg) {
+  const auto& use_udp = control->object_storage()->set_c_str_bool("trackers.use_udp", arg);
+
+  if (use_udp.as_value() == 0) {
+    for (auto download : *control->core()->download_list())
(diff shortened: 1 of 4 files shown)


rakshasa:
https://github.com/rakshasa/rtorrent/pull/1775

PR #1777, opened 2026-05-17
SCGI: use global SocketManager with per-category limit

### LINKED ISSUE #1776: SCGI socket pool should be independent from BitTorrent peer connections pool
## Problem

https://github.com/rakshasa/rtorrent/commit/71ac256cb5f0b4dc007413aa95a9137b3fbbc735

SCGI RPC connections and BitTorrent peer connections share the same global `SocketManager` pool, both gated by `network.max_open_sockets`. Under normal torrent load with many peers, the pool fills up and `SocketManager::open_event_or_cleanup` rejects SCGI accept by silently closing the newly accepted fd via `cleanup(false)`. From the client side this manifests as the connection being immediately dropped — **RPC becomes unavailable exactly when torrent activity is high**, which is precisely when you need monitoring and management the most.

SCGI already has its own concurrency limit `SCgi::max_tasks = 100`, but this only guards task slot assignment and comes **after** the socket pool check. A full socket pool kills the connection before the task slot check ever runs.

## Root cause

Both go through the same `SocketManager::open_event_or_cleanup` with the shared `m_max_size`:

- **SCGI accept** — `src/rpc/scgi.cc:186`
```cpp
bool result = torrent::runtime::socket_manager()->open_event_or_cleanup(
    m_current->get(), open_func, cleanup_func);
```

- **BT handshake** — `libtorrent/src/protocol/handshake_manager.cc:195`
```cpp
runtime::socket_manager()->open_event_or_cleanup(handshake.get(), open_func, cleanup_func);
```

- **Shared limit check** — `libtorrent/src/torrent/runtime/socket_manager.cc:105`
```cpp
if (m_managed_size + m_unmanaged_size >= m_max_size) {
    cleanup(false);  // closes fd silently for SCGI
    return false;
}
```

The global `m_max_size` is computed once at startup from total fd budget minus BT-specific reserves (`torrent.cc:141`), with no separate allocation for SCGI:
```cpp
runtime::socket_manager()->set_max_size(max_open - max_files - max_http_connections - reserved);
```

## Impact

- RPC becomes unavailable under heavy torrent load — management tools (ruTorrent, flood, custom scripts) cannot connect
- The failure mode is silent: the connection is accepted by the kernel but immediately closed, giving no meaningful error to the client
- Two unrelated subsystems (RPC management vs. P2P data transfer) compete for the same pool, causing a denial-of-service on RPC when BT peers are numerous

## Possible approaches

1. Give SCGI listener and SCGI tasks a separate `SocketManager` instance or an independent counter, so they are only bounded by `SCgi::max_tasks` (plus 1 for the listener) rather than the global BT pool
2. Let SCGI connections bypass the `SocketManager` global limit entirely (SCGI has its own `max_tasks` boundary already)
3. Add per-category quota support in `SocketManager` (e.g., tag SCGI vs BT events so they count against separate sub-limits)

trim21:
SCGI connections now use the global `torrent::runtime::socket_manager()` 
with `category_scgi` instead of a dedicated private `SocketManager`.

SCGI capacity is bounded by its own category max (set to `max_tasks + 1`) 
via `set_category_max_size()`, preventing it from starving the peer 
connection pool within the global fd budget.

This relies on the per-category allocation limits added to SocketManager 
in rakshasa/libtorrent#724.

Closes #1776

### DIFF
--- a/src/rpc/scgi.cc
+++ b/src/rpc/scgi.cc
@@ -11,6 +11,7 @@
 #include <torrent/net/poll.h>
 #include <torrent/net/socket_address.h>
 #include <torrent/runtime/socket_manager.h>
+#include <torrent/utils/log.h>
 
 #include "control.h"
 #include "globals.h"
@@ -22,6 +23,14 @@
 namespace rpc {
 
 SCgi::SCgi() {
+  auto category_max = max_tasks + 1;
+
+  torrent::runtime::socket_manager()->set_category_max_size(torrent::runtime::SocketManager::category_scgi, category_max);
+
+  if (category_max >= torrent::runtime::socket_manager()->max_size())
+    lt_log_print(torrent::LOG_WARN, "SCGI category max %u >= global max %u, SCGI could starve peer connections",
+                 category_max, torrent::runtime::socket_manager()->max_size());
+
   std::generate(m_tasks.begin(), m_tasks.end(), []() { return std::make_unique<SCgiTask>(); });
 
   m_current = m_tasks.begin();
@@ -48,7 +57,7 @@ SCgi::open_port(sockaddr* sa, unsigned int length, bool dont_route) {
 
   open(reinterpret_cast<sockaddr*>(sa), length);
 
-  torrent::runtime::socket_manager()->register_event_or_throw(this, []() {});
+  torrent::runtime::socket_manager()->register_event_or_throw(this, torrent::runtime::SocketManager::category_scgi, []() {});
 }
 
 void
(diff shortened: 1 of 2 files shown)


rakshasa:
SocketManager is global for all fd's for a reason, and this needs to be fixed with changes to how SCGI interacts with SocketManager.

Making a separate SocketManager defeats the whole purpose of it.

rakshasa:
Possible solution is to add categories of allocated open sockets to SocketManager, and give peer connections (etc) the main allocation, http connections another, and scgi, etc.

trim21:
looks like some files are not clang-formatted...

rakshasa:
Please don't reformat code, it adds unnecessary junk to PRs and quite often clang formatting is does things I don't want it to do.

src/rpc/scgi.cc
```diff
@@ -22,6 +23,14 @@
 namespace rpc {
 
 SCgi::SCgi() {
+  auto category_max = max_tasks + 1;
+
+  torrent::runtime::socket_manager()->set_category_max_size(torrent::runtime::SocketManager::category_scgi, category_max);
+
+  if (category_max >= torrent::runtime::socket_manager()->max_size())
+    lt_log_print(torrent::LOG_WARN, "SCGI category max %u >= global max %u, SCGI could starve peer connections",
+                 category_max, torrent::runtime::socket_manager()->max_size());
```
rakshasa: Seems we should just adjust `max_tasks` to be the size of the category.

PR #1790, opened 2026-05-29
feat: auto-detect LuaJIT with --with-lua

trim21:
The `ax_lua.m4` macro already has full LuaJIT support (interpreter, header, and library detection), but it was explicitly disabled in `configure.ac` with `AM_CONDITIONAL([LUAJIT], [false])` because the wiring wasn't done yet (commit 8805100).

This PR completes that work by making `--with-lua` auto-detect LuaJIT:

1. **`scripts/checks.m4`**: Before calling `AX_PROG_LUA`, probe for `luajit` in `PATH`. If found, set `AM_CONDITIONAL([LUAJIT], [true])` so the downstream macros search the correct LuaJIT paths (`/usr/include/luajit-VERSION/`, `luajit-5.1` lib, etc.).

2. **`scripts/ax_lua.m4`**: Append `luajit` to the non-LUAJIT interpreter search list as a fallback, and remove hardcoded beta version entries (`luajit-2.1.0-beta3`, etc.) that are no longer relevant.

3. **`configure.ac`**: Update comment to reflect the new auto-detection behavior.

## Behavior

- `./configure --with-lua` on a system with LuaJIT → uses LuaJIT automatically
- `./configure --with-lua` on a system with only PUC Lua → uses PUC Lua
- `LUA=/path/to/lua5.4 ./configure --with-lua` → forces PUC Lua even if LuaJIT is installed

## Tested

Verified on Debian with `libluajit-5.1-dev` installed:
- `configure` correctly finds `/usr/bin/luajit`, library `-lluajit-5.1`, and headers in `/usr/include/luajit-2.1`
- `src/rpc/lua.cc` compiles cleanly against LuaJIT headers

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -55,7 +55,7 @@ AC_LANG_PUSH(C++)
 TORRENT_WITH_XMLRPC_C
 AC_LANG_POP(C++)
 
-dnl We don't have LuaJIT support, and the script breaks if not disabled.
+dnl Default to false; TORRENT_WITH_LUA will set it to true if LuaJIT is found.
 AM_CONDITIONAL([LUAJIT], [false])
 
 TORRENT_WITH_LUA
--- a/scripts/ax_lua.m4
+++ b/scripts/ax_lua.m4
@@ -207,8 +207,8 @@ AC_DEFUN([AX_PROG_LUA],
 
   dnl Find a Lua interpreter.
   AM_COND_IF([LUAJIT],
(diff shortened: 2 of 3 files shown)


trim21:
（I would really recommend meson, the autotools is getting more and more complex)

rakshasa:
https://github.com/rakshasa/rtorrent/compare/master...feature/base64-support

Contains a fix, use that as a base to fix LuaJIT support.

PR #1793, opened 2026-05-29
Added support base64 support for non-UTF8 strings.

rakshasa:


xirvik:
This seems to break builds

 return xmlrpc_base64_new(env, object.as_string().c_str(), object.as_string().size());

rpc/xmlrpc_c.cc: error: invalid conversion from 'size_type' {aka 'long unsigned int'}
                   to 'const unsigned char*' [-fpermissive]

rakshasa:
Fixed in PR, also adding unit-tests to check both xmlrpc-c and tinyxml2 compiles.

PR #1796, opened 2026-06-03
fix: set d.message on hash check failure

trim21:
## Summary

- Populate `d.message` with the specific error message when a hash check fails, instead of leaving it empty.

## Motivation

When rtorrent finishes downloading and the final hash check encounters an I/O error, `d.hashing_failed` is set to `true` but `d.message` is left empty. This makes it impossible to debug the root cause of the failure.

## Changes

In `DownloadList::hash_done()`, after setting `hash_failed(true)`, retrieve the error message from the hash checker via the new `Download::hash_error_message()` API and set it as `d.message`.

Also fix `Manager::receive_hashing_changed()` where `set_hash_failed(true)` was called without setting `d.message` when catching `local_error` during hash check (e.g. "too many open files").

## Dependency

Requires [libtorrent#747](https://github.com/rakshasa/libtorrent/pull/747) which adds `Download::hash_error_message()`.

### DIFF
--- a/src/core/download_list.cc
+++ b/src/core/download_list.cc
@@ -492,7 +492,11 @@ DownloadList::hash_done(Download* download) {
 
   if (!download->is_hash_checked()) {
     download->set_hash_failed(true);
-    
+
+    auto& msg = download->download()->hash_error_message();
+    if (!msg.empty())
+      download->set_message(msg);
+
     DL_TRIGGER_EVENT(download, "event.download.hash_failed");
     return;
   }
--- a/src/core/manager.cc
+++ b/src/core/manager.cc
@@ -513,6 +513,7 @@ Manager::receive_hashing_changed() {
 
       } else {
         (*itr)->set_hash_failed(true);
+        (*itr)->set_message("Hashing failed: " + std::string(e.what()));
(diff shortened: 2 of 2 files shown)


src/core/download_list.cc
```diff
@@ -492,7 +492,11 @@ DownloadList::hash_done(Download* download) {
 
   if (!download->is_hash_checked()) {
     download->set_hash_failed(true);
-    
+
+    int err = download->download()->hash_error_number();
+    if (err != 0)
+      download->set_message("Hash check failed: " + std::string(std::strerror(err)));
```
rakshasa: This doesn't set any message if the error number is zero for some reason.

So just pass the error number directly to std::strerror, it will show `Success` in that case.

trim21: sorry, this is not ready for review yet. I forget to set it to draft. I'm still testing this locally

trim21:
this should work now, I see a clear error message `Hashing failed: Hash check I/O error at chunk 0: Cannot allocate memory | ` now

trim21:
@rakshasa you didn't merge this one……

PR #1801, opened 2026-06-04
rpc: fix --with-xmlrpc-c backend build (flag name + base64 args)

xirvik:
The XMLRPC-C backend (src/rpc/xmlrpc_c.cc) has not compiled since the base64-for-non-UTF8 work landed. object_to_xmlrpc() has three errors, none caught by CI: the unit-tests workflow runs a bare ./configure with neither --with-xmlrpc-c nor --with-xmlrpc-tinyxml2 and installs no xmlrpc-c headers, so this translation unit is never built upstream.

1. torrent::Object::flag_binary does not exist; the libtorrent flag is flag_as_binary (0x800000, "Treat as binary data"). Use it.

2-3. xmlrpc_base64_new() takes (xmlrpc_env*, size_t length,
   const unsigned char* value), but both calls pass (value, length), which
   fails to compile: invalid conversion from size_type to const unsigned char*.
   Swap the arguments to match the library.

Builds cleanly with --with-xmlrpc-c against current libtorrent master.

### DIFF
--- a/src/rpc/xmlrpc_c.cc
+++ b/src/rpc/xmlrpc_c.cc
@@ -278,7 +278,7 @@ object_to_xmlrpc(xmlrpc_env* env, const torrent::Object& object) {
 
   case torrent::Object::TYPE_STRING:
   {
-    if (object.flags() & torrent::Object::flag_binary) {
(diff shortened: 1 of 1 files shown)


rakshasa:
Already fixed.

PR #1809, opened 2026-06-07
add missing maybe_unuseds

fffe:
there are a couple of variables that the without-ncurses macros don't touch, that spam the build logs

this updates them to match the other functions.


src/display/canvas.h
```diff
@@ -144,7 +144,8 @@ Canvas::get_x() {
 
 inline unsigned int
 Canvas::get_y() {
-  int x, y;
+  int y;
+  [[maybe_unused]] int x;
```
rakshasa: Why not just add the attribute for both, it's cleaner.

fffe: because the other variable is always used, but if you prefer I can resubmit the patch with those changes (and cleaning up the others)

I forget, did you want me to force pushes onto PRs instead of adding update commits? (and dumb git question, is there a reason you don't squash merge PRs with multiple commits?)

rakshasa: It makes the code cleaner in this case.

Also for small updates like this force-push is better, and I prefer to not squash external PRs since I don't know what kind of attribution the contributor prefers.

fffe:
alright, they're now all `[[maybe_unused]] int x, y;`

also adding a related configure fix: the test for pkg-config is hidden in scripts/ax_with_curses.m4, and if --without-ncurses is given, that branch is skipped. Unless the PKG_CONFIG env var is set, pkg-config won't be detected, and the zlib, libtorrent etc. checks will fail.

this adds an explicit test for pkg-config with the other tool checks.

fffe:
hmm... I'm not sure why `git push origin -f sometimes_unused` didn't work, but I don't care if you squash this.

PR #1820, opened 2026-06-10
fix: replace reinterpret_cast UB in command_base with aligned placement new

### LINKED ISSUE #1818: `command_base::_pod()` performs undefined behavior via `reinterpret_cast` between different `std::function` specializations — causes crashes in debug/UBSan builds
### Location

`rtorrent/src/rpc/command.h:147`

```cpp
template <typename tmpl> tmpl& _pod() { return reinterpret_cast<tmpl&>(t_pod); }
```

### The UB

`t_pod` is declared as `base_function` = `std::function<torrent::Object(target_type, const torrent::Object&)>`.

But `_pod<Func>()` is called with **different** `std::function` specializations, such as:

| Macro | Stored `std::function` type | Retrieved `std::function` type |
|---|---|---|
| `CMD2_ANY_VALUE` | `std::function<Object(target_type, const Object&)>` | `std::function<Object(target_type, const int64_t&)>` |
| `CMD2_ANY_STRING` | `std::function<Object(target_type, const Object&)>` | `std::function<Object(target_type, const std::string&)>` |
| `CMD2_ANY_LIST` | `std::function<Object(target_type, const Object&)>` | `std::function<Object(target_type, const list_type&)>` |
| `CMD2_DL` | `std::function<Object(target_type, const Object&)>` | `std::function<Object(core::Download*, const Object&)>` |

`std::function<R(A)>` and `std::function<R(B)>` are **distinct, unrelated types**. Accessing one through a `reinterpret_cast` to another violates C++ strict aliasing and lifetime rules. This is undefined behavior per [\[basic.lval\]/11](https://eel.is/c++draft/basic.lval#11).

The code itself acknowledges this in a comment at `command.h:142`:
> "The std::function object in GCC is castable between types ... yet what the C++0x standard will say about this I have no idea atm."

### Real-world impact

In **debug builds** (e.g. `-O0 -g`), and especially when compiled with:
- libc++ (via zig cc or clang)
- UBSan (`-fsanitize=undefined`)
- AddressSanitizer (`-fsanitize=address`)

this manifests as:
- **SIGSEGV at startup** in `std::function` internal destructor during `system_listMethods()` (which iterates `CommandMap`, copying each entry)
- **Misaligned access crashes** when `std::function` requires 16-byte alignment but `t_pod` is stored in a `char[]` union (1-byte alignment)
- **`double free or corruption`** at shutdown when destructors run on mismatched `std::function` types

On GCC with libstdc++ in release builds, it "works" because all `std::function` specializations happen to share the same internal small-buffer layout and vtable structure. But this is an implementation detail and not guaranteed.

### The call chain that crashes

```
system_listMethods()                     → iterates rpc::commands map by value
  pair::~pair()                          → copies command_map_data_type, destroying copy
    command_map_data_type::~()           → destroys command_base member
      command_base::~command_base()      → calls _pod<base_function>().~base_function()
        std::function::~function()       → SIGSEGV if t_pod contains a different specialization
```

The copy constructor (`command.h:119`) also reads `t_pod` as `base_function` via `_pod<base_function>()` to construct the copy, which reads garbage if a different `std::function` type was stored.

### Suggested fix

Replace the `reinterpret_cast`-based type erasure with proper C++ type erasure, e.g.:

1. Store a `std::any` or `std::variant<all_possible_function_types...>` instead of raw `reinterpret_cast`
2. Or unify all slot signatures to `torrent::Object(target_type, const torrent::Object&)` and perform argument conversion (`.as_value()/.as_string()/.as_list()`) at the call boundary instead of through `reinterpret_cast`

trim21:
## Summary

Replace `reinterpret_cast`-based type erasure in `command_base::_pod()` with `alignas` inline storage + typed copy/destroy helpers, eliminating undefined behavior when storing different `std::function` specializations.

## Problem

`command_base::t_pod` is a `base_function`, but `set_function<T>()` stores different `std::function` specializations via `reinterpret_cast`. The copy ctor and destructor always treat the stored value as `base_function`, which is UB when a different specialization was stored.

`std::function<A>` and `std::function<B>` are distinct, unrelated types. Accessing one through `reinterpret_cast` to another violates strict aliasing ([basic.lval]/11).

## Fix

Replace `union { base_function t_pod; }` with aligned `char` buffer + typed lifecycle helpers:

| Component | Before (UB) | After (safe) |
|---|---|---|
| Storage | `union { base_function t_pod; }` | `alignas(max_align_t) char t_pod[...]` |
| `_pod<T>()` | `reinterpret_cast<T&>` on wrong union member | `reinterpret_cast<T&>` — safe, `T` was placement-newed there |
| `set_function<T>(s)` | writes through wrong union member | `::new (t_pod) T(...)` + stores typed copy/destroy helpers |
| Copy ctor | `_pod<base_function>()` — wrong type | Calls `m_copy_helper` — knows actual type |
| Destructor | `~base_function()` — wrong type | Calls `m_dest_helper` — destroys correct type |

Zero runtime overhead — no RTTI, no type index checks, no heap allocation.

Closes #1818

### DIFF
--- a/src/rpc/command.h
+++ b/src/rpc/command.h
@@ -3,6 +3,7 @@
 
 #include <functional>
 #include <limits>
+#include <new>
 #include <torrent/common.h>
 #include <torrent/object.h>
 #include <torrent/data/file_list_iterator.h>
@@ -28,7 +29,6 @@ struct target_wrapper<void> {
   typedef no_type*    cleaned_type;
 };
 
-// Since c++0x isn't out yet...
 template <typename T1, typename T2, typename T3>
 struct rt_triple : private std::pair<T1, T2> {
   typedef std::pair<T1, T2> base_type;
@@ -55,9 +55,6 @@ struct rt_triple : private std::pair<T1, T2> {
     base_type(src.first, src.second), third(src.third) {}
 };
 
-// Since it gets used so many places we might as well put it in the
-// rpc namespace.
-//typedef std::pair<int, void*> target_type;
 typedef rt_triple<int, void*, void*> target_type;
 
 class command_base;
@@ -114,10 +111,48 @@ public:
     char buffer[sizeof(torrent::Object) * max_arguments];
   };
 
-  command_base() { new (&_pod<base_function>()) base_function(); }
-  command_base(const command_base& src) { new (&_pod<base_function>()) base_function(src._pod<base_function>()); }
+  command_base() : m_copy_helper(nullptr), m_dest_helper(nullptr) {}
 
-  ~command_base() { _pod<base_function>().~base_function(); }
+  command_base(const command_base& src) {
+    m_copy_helper = src.m_copy_helper;
+    m_dest_helper = src.m_dest_helper;
+    if (src.m_copy_helper)
+      src.m_copy_helper(t_pod, src.t_pod);
+  }
+
+  command_base& operator=(const command_base& src) {
+    if (this != &src) {
+      if (m_dest_helper) m_dest_helper(t_pod);
+      m_copy_helper = src.m_copy_helper;
+      m_dest_helper = src.m_dest_helper;
+      if (src.m_copy_helper)
+        src.m_copy_helper(t_pod, src.t_pod);
+    }
+    return *this;
+  }
+
+  command_base(command_base&& src) noexcept {
+    m_copy_helper = src.m_copy_helper;
+    m_dest_helper = src.m_dest_helper;
+    if (src.m_copy_helper)
+      src.m_copy_helper(t_pod, src.t_pod);
+  }
+
+  command_base& operator=(command_base&& src) noexcept {
+    if (this != &src) {
+      if (m_dest_helper) m_dest_helper(t_pod);
+      m_copy_helper = src.m_copy_helper;
+      m_dest_helper = src.m_dest_helper;
+      if (src.m_copy_helper)
+        src.m_copy_helper(t_pod, src.t_pod);
+    }
+    return *this;
+  }
+
+  ~command_base() {
+    if (m_dest_helper)
+      m_dest_helper(t_pod);
+  }
 
   static torrent::Object* argument(unsigned int index) { return current_stack.begin() + index; }
   static torrent::Object& argument_ref(unsigned int index) { return *(current_stack.begin() + index); }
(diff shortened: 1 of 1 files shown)


rakshasa:
This adds type data to the POD and results in less efficient code.

Use either memcpy (which gets optimized away):

```
template <typename tmpl> 
tmpl _pod() const { 
    static_assert(std::is_trivially_copyable_v<tmpl>);
    tmpl result;
    std::memcpy(&result, &t_pod, sizeof(tmpl));
    return result; 
}
```

Or std::bit_cast:

```
using GenericFuncPtr = void(*)();
GenericFuncPtr t_pod;

template <typename tmpl> 
tmpl _pod() const { 
    static_assert(sizeof(tmpl) == sizeof(t_pod), "Size mismatch for bit_cast");
    static_assert(std::is_trivially_copyable_v<tmpl>, "Target must be trivially copyable");

    return std::bit_cast<tmpl>(t_pod); 
}
```

rakshasa:
Variant objects still contains type information and it gets checked, so it adds both memory and instructions.

trim21:
I think both `std::bit_cast` and `memcpy` won't work, you check it with `static_assert(std::is_trivially_copyable_v<tmpl>)` but `tmpl` here is `std::function` and it's not trivially copyable

variant do still have both memory and instructions but there is no UB and more efficiency than `std::any`.

Otherwise, we will need to rewrite every handler to make it zero runtime overhead and no UB.

trim21:
for clarify, I'm ok with UB in cxx if it works fine on the target platform, but this case it break debug build for debugging so I hope we can fix this

rakshasa:
The above examples I posted _do_ fix the UB in a standard-compliant way, as they avoid memory aliasing.

Since we use C++20 I'd prefer bit_cast.

trim21:
> The above examples I posted _do_ fix the UB in a standard-compliant way, as they avoid memory aliasing.
> 
> Since we use C++20 I'd prefer bit_cast.

sorry, I don't understand, `static_assert(std::is_trivially_copyable_v<T>)` doesn't compile, here T is `std::function` and it's not trivially copyable, how would you want to use bit_cast here? doing bit_cast and memcpy are u

rakshasa:
```
class command_base {
public:
  using copy_fn_t = void(*)(void* dest, const void* src);
  using dest_fn_t = void(*)(void* ptr);

  command_base() : _copy_helper(nullptr), _dest_helper(nullptr) {}

  command_base(const command_base& src) {
    _copy_helper = src._copy_helper;
    _dest_helper = src._dest_helper;
    if (src._copy_helper) {
      src._copy_helper(this->t_pod, src.t_pod);
    }
  }

  command_base& operator=(const command_base& src) {
    if (this != &src) {
      if (_dest_helper) _dest_helper(t_pod);
      _copy_helper = src._copy_helper;
      _dest_helper = src._dest_helper;
      if (src._copy_helper) {
        src._copy_helper(this->t_pod, src.t_pod);
      }
    }
    return *this;
  }

  ~command_base() { 
    if (_dest_helper) {
      _dest_helper(t_pod); 
    }
  }

  template <typename T>
  void set_function(T s, [[maybe_unused]] int value = command_base_is_valid<T>::value) {
    static_assert(sizeof(T) <= sizeof(t_pod), "t_pod storage overflow!");
    static_assert(alignof(t_pod) % alignof(T) == 0, "t_pod alignment violation!");

    if (_dest_helper) _dest_helper(t_pod);

    ::new (t_pod) T(std::move(s));

    _copy_helper = [](void* dest, const void* src) {
      ::new (dest) T(*reinterpret_cast<const T*>(src));
    };
    _dest_helper = [](void* ptr) {
      reinterpret_cast<T*>(ptr)->~T();
    };
  }

  template <command_base_call_type T>
  void set_function_2(typename command_base_is_type<T>::type s, [[maybe_unused]] int value = command_base_is_valid<typename command_base_is_type<T>::type>::value) {
    set_function<typename command_base_is_type<T>::type>(std::move(s));
  }

  template <typename tmpl> tmpl& _pod() { return reinterpret_cast<tmpl&>(t_pod); }
  template <typename tmpl> const tmpl& _pod() const { return reinterpret_cast<const tmpl&>(t_pod); }

protected:
  alignas(std::max_align_t) char t_pod[sizeof(base_function)];

  copy_fn_t _copy_helper;
  dest_fn_t _dest_helper;
};
```

This should be better imo.

It avoids any extra instructions when accessing the POD, while keeping it correct when ctor/dtor/copying.

PR #1828, opened 2026-06-14
main: register deprecated-command aliases unconditionally

xirvik:
0.16.14 (commit 7537781) moved a batch of deprecated-command redirects (execute2, schedule2, bind, ip, port_range, dht, port_random, network.http.max_open, max_memory_usage, encoding_list, ...) inside an 'if (method.use_deprecated == 1)' guard in main(). That value is read during command registration, BEFORE parse_config_file() runs, and defaults to false; it can only be set early via the -D flag. So 'method.use_deprecated.set = true' in the rc has no effect on registration, and these aliases do not exist when the rc (or a front-end) uses them.

Net effect: 'rtorrent -o import=<file>' aborts at config-parse on any rc using a deprecated alias on a default invocation - a regression from <= 0.16.13, where the same block was unconditional (the guard was commented out). Register the aliases unconditionally via a plain scope block; they are cheap redirect entries and gating their registration on a value the config itself is meant to set is a chicken-and-egg bug.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -353,7 +353,13 @@ main(int argc, char** argv) {
     //   Allow for use in config files, etc, just don't export it.
     // }
 
-    if (rpc::call_command_value("method.use_deprecated") == 1) {
+    // Deprecated-command aliases are registered unconditionally. Gating their
+    // registration on method.use_deprecated was a chicken-and-egg bug: that value is
+    // normally enabled from the config file (or by ruTorrent), but the config is parsed
+    // (parse_config_file, below) AFTER this point -- so config-based enabling never took
+    // effect; only the -D command-line flag did. Registering them unconditionally keeps
+    // them usable via `rtorrent -o import=<rc>` and ruTorrent, as in <= 0.16.13.
+    {
       CMD2_REDIRECT("execute2",         "execute");
       CMD2_REDIRECT("schedule2",        "schedule");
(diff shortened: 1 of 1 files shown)


rakshasa:
The point of moving them there is to force users to update their rc files as they are going to be removed in the future.

Many of those commands have been replaced by new commands many years ago, or they are no longer supported.

The option to enable them is a way for users who don't want to update rc files to do so later, however if they are just enabled by default then there's no incentive for users to fix their rc-file.

xirvik:
> The point of moving them there is to force users to update their rc files as they are going to be removed in the future.

I see.  It's a reasonable point of view. But since this breaks existing deployments, maybe it shouldn't be a minor version bump? 0.16.14 is not compatible with the same config files as 0.16.13 because of this.

Also ruTorrent itself is choking on this change now, but let's say that's not a big deal since I can update it myself.

rakshasa:
I'm essentially cleaning house before an eventual 1.0 (or rather 1.1) release, so 0.16.x will have some changes that will break configs.

xirvik:
> I'm essentially cleaning house before an eventual 1.0 (or rather 1.1) release, so 0.16.x will have some changes that will break configs.

OK. I released a new ruTorrent supporting 0.16.14. If something is still broken I'll fix promptly.

Closing this PR.

PR #1836, opened 2026-06-18
command: guard d.bitfield against an unallocated (null) bitfield

xirvik:
A closed or never-opened download keeps Bitfield::size_bytes() > 0 but frees the data buffer (empty() == m_data == nullptr). d.bitfield passed *bitfield() to transform_to_hex_str(), which iterates [begin(), end()) == [nullptr, nullptr + size_bytes()) and dereferences null -> SIGSEGV in torrent::utils::transform_to_hex_str (string_manip.h:321). Reachable from any XML-RPC client polling d.bitfield via d.multicall2 (ruTorrent, pymedusa); backend-independent (seen via both xmlrpc-c and tinyxml2). Surfaced with the 0.16.13 callback/bitfield-lifecycle changes; 0.16.12 was the last working version per the reporter. Fix: return an empty string for an unallocated bitfield instead of hexing a null range.

More context:
https://github.com/Novik/ruTorrent/issues/3065#issuecomment-4743730314

### DIFF
--- a/src/command_download.cc
+++ b/src/command_download.cc
@@ -646,7 +646,15 @@ initialize_command_download() {
   CMD2_DL("d.hash",                           [](auto* download, auto) { return torrent::utils::transform_to_hex_str(download->info()->hash()); });
   CMD2_DL("d.local_id",                       [](auto* download, auto) { return torrent::utils::transform_to_hex_str(download->info()->local_id()); });
   CMD2_DL("d.local_id_html",                  [](auto* download, auto) { return torrent::utils::copy_escape_html_str(download->info()->local_id()); });
-  CMD2_DL("d.bitfield",                       [](auto* download, auto) { return torrent::utils::transform_to_hex_str(*download->download()->file_list()->bitfield()); });
+  CMD2_DL("d.bitfield",                       [](auto* download, auto) {
+      // A closed/unopened download has a Bitfield with size_bytes() > 0 but a freed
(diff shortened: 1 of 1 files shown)


mirolm:
Should it be better "transform_to_hex_str" to just return empty string if the input is empty?

xirvik:
> Should it be better "transform_to_hex_str" to just return empty string if the input is empty?

The input isn't actually empty here — it's a null pointer with a nonzero length (begin=0x0, end=0x80), so an "is it empty" check wouldn't trigger; the deref happens because begin != end. And folding this into the generic transform_to_hex_str would make a low-level helper silently swallow null-data inputs, masking real bugs elsewhere — whereas guarding d.bitfield  expresses the actual invariant (a closed download has no bitfield).

rakshasa:
This isn't really the correct solution, Bitfield is supposed to act like std containers so `begin() == end()` should be fixed to be true.

rakshasa:
Looks like the correct solution is to check for `m_data == nullptr` in end(), as a Bitfield with `m_set == 0 || m_set == m_size` can be empty.

xirvik:
> Looks like the correct solution is to check for `m_data == nullptr` in end(), as a Bitfield with `m_set == 0 || m_set == m_size` can be empty.

https://github.com/rakshasa/libtorrent/pull/786

Closing this one.

PR #1837, opened 2026-06-18
input: survive controlling-terminal hangup instead of aborting the client

xirvik:
Since the 0.16.13 callback/poll rework, an EPOLLERR/EPOLLHUP on the terminal input fd (stdin) makes Poll::process() throw torrent::internal_error ("received error event for event not in error: name:input fd:0") and the whole client dies. epoll always delivers EPOLLERR/EPOLLHUP regardless of registration, but InputEvent only registered for read, so the error event was unexpected -> fatal. Hit in production (rtorrent runs under a pty) and reproducible under XML-RPC load.

Fix mirrors the SCgiTask error pattern: register the input event for error events (insert_error) so the hangup is dispatched to event_error(), and have event_error() drop stdin from the poll set (idempotent via the stored Poll*) so rtorrent keeps running headless rather than aborting. Differential repro: stock logs the internal_error to crash.log and dies; patched never logs it and survives (2.2M+ XML-RPC requests).

### DIFF
--- a/src/input/input_event.cc
+++ b/src/input/input_event.cc
@@ -13,18 +13,11 @@ void
 InputEvent::insert() {
   torrent::this_thread::poll()->open(this);
   torrent::this_thread::poll()->insert_read(this);
-  // EPOLLERR/EPOLLHUP is always delivered by epoll regardless of registration;
-  // register for it so a controlling-terminal/pty hangup on stdin is dispatched
-  // to event_error() instead of aborting the whole client (see event_error()).
   torrent::this_thread::poll()->insert_error(this);
 }
 
 void
 InputEvent::remove() {
-  // The file descriptor doubles as the open-state guard, mirroring SCgiTask.
-  // event_error() may already have dropped stdin from the poll set and cleared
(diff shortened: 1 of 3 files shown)


src/input/input_event.cc
```diff
@@ -32,6 +45,13 @@ InputEvent::event_write() {
 
 void
 InputEvent::event_error() {
+  // The controlling terminal/pty hung up (EPOLLERR on stdin). Stop polling it
+  // rather than letting Poll::process() throw an internal_error and kill the
+  // daemon -- a regression introduced with the 0.16.13 callback/poll rework.
+  // rtorrent keeps running with no keyboard input; fd is just dropped from the
+  // poll set (idempotent with remove()).
+  if (m_poll != nullptr)
+    remove(m_poll);
```
rakshasa: There's no need for an `m_poll` object in InputEvent, use `torrent::this_thread::poll()` instead, and just remove it no need to guard against further event_error().

And poll isn't really needed to be passed anymore above either.

rakshasa: BTW, use remove_and_close().

xirvik: Done.

PR #1844, opened 2026-06-24
Replace rak::regex with POSIX fnmatch

trim21:
## Summary

Replace the custom `rak::regex` glob matcher with the standard POSIX `fnmatch()` from `<fnmatch.h>`.

### Motivation

- `rak::regex` was a hand-rolled implementation supporting only `*` wildcards
- The original code comment itself says: *"Replace with TR1's regex when that becomes widely available"*
- `fnmatch()` is the POSIX-standard way to do shell-style glob pattern matching
- It supports `*`, `?`, and `[...]` character classes — a strict superset of the old behavior

### Changes

| File | Change |
|---|---|
| `src/core/manager.cc` | Replace `rak::regex` functor with `fnmatch()` call in directory entry filtering |
| `src/command_download.cc` | Replace `std::vector<rak::regex>` with `std::vector<std::string>` + `fnmatch()` for file path matching |
| `Makefile.am` | Remove `rak/regex.h` from `EXTRA_DIST` |
| `rak/regex.h` | Deleted (110 lines removed) |

### Compatibility

- `fnmatch(pattern, text, 0)` is called **without** `FNM_PATHNAME`, so `*` can still match `/` — preserving the original behavior
- The dotfile filtering logic (`update_hide_dot`) is preserved independently of the pattern matching


rakshasa:
Why not C++ regex?

trim21:
> Why not C++ regex?

too slow, and rak::regex is not actually regex but glob

PR #1846, opened 2026-06-25
GHA CI: enable libtorrent build flag matrix

kingomarwashere:
## Summary

- Restructures the `ubuntu-base` libtorrent build job to use a `libt_configure_flags` matrix variable
- Adds an `--enable-openssl` variant alongside the default build, exercising the flag-passing infrastructure
- Artifact names are made unique per matrix entry so downstream jobs can reference the correct build
- Mirrors the structural approach taken in [qbittorrent/qBittorrent#24564](https://github.com/qbittorrent/qBittorrent/pull/24564)

This paves the way for future WebTorrent related development: when rakshasa/libtorrent gains `--enable-webtorrent` support, adding it to CI will be a one-line matrix entry change.

> **Note:** rakshasa/libtorrent uses an autotools build system (`./configure`) rather than CMake, so the flag variable is `libt_configure_flags` rather than CMake `-D` flags. The pattern is otherwise identical to the qBittorrent approach.

## Test plan

- [ ] `ubuntu-base` builds libtorrent twice: default and `--enable-openssl`
- [ ] `unit-tests-default` and `unit-tests-variants` continue to pass using the `default` artifact

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -6,6 +6,12 @@ jobs:
 
   ubuntu-base:
     runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        libtorrent:
+          - libt_configure_flags: "--enable-openssl"
+          - libt_configure_flags: ""
     steps:
     - name: Update Packages
       run: |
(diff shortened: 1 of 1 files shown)


rakshasa:
Why is this PR necessary, is someone implementing webtorrent support?

kingomarwashere:
I'm taking a run at implementing it. I just wanted to get your attention to see if you'd accept an implementation before we invest a couple weeks @rakshasa 
you in or are you in?

rakshasa:
It always depends on the quality of the implementation, and how it is structured.

kingomarwashere:
Great :) How would you like it to be structured @rakshasa any specific do's and do-nots?

Issue #1848, opened 2026-06-26
execute.*.bg broken since dbe7997 (the posix_spawn change)


fffe:
dbe7997 dropped the code supporting `flag_background` for the execute methods, causing execute.*.bg calls that trigger RPC requests to deadlock.

It also dropped the fd close loop in favor of `POSIX_SPAWN_CLOEXEC_DEFAULT` -- despite the comment, that's macOS-only, so I think most systems are leaking fds on exec. `posix_spawn_file_actions_addclosefrom_np` exists on recent linux (glibc) + freebsd, but not musl. The proper fix is probably adding CLOEXEC everywhere, but that's a bigger change so I ignored it for now.

fffe:
background jobs were fixed in 0.16.6, opened https://github.com/rakshasa/libtorrent/issues/802 to track O_CLOEXEC and fds leaking on musl + non-freebsd bsds.

fffe:
I was wrong, this is still broken on 0.16.16. 786f1234d2f948a2e8eacd86072188b313fb5bfa didn't fix it.

rakshasa:
In what way is it broken?

```
# Executed in background, sticks around after rtorrent shuts down
execute.throw.bg=tail,-f,/dev/null

# Gone immediately, pid harvested
execute.throw.bg=tail,/dev/null 

# Hangs rtorrent
execute.throw=tail,-f,/dev/null
```

These are the expected behaviors, on a macos.

fffe:
With the exec-order branch applied, my move on complete script goes defunct in a linux container (execute.nothrow.bg on finished + schedule)

rakshasa:
Might be that it relied on timing issues that no longer apply?

We're no longer doing double-forking, so it's faster now. (and the main process gets frozen until posix_spawn is done, so it makes any timing issues even more likely to trigger)

rakshasa:
Using exec-order, `sleep,20` in background got cleaned up after 20 seconds.

So on macos it works.

fffe:
Maybe ignore this one for now and let me dig a bit more.

My test environment comes through github actions, and I was getting layer caching that... shouldn't have been possible.

I'll ping you if I have more details, or close it if I don't.

fffe:
Yeah, it was something screwy with container cache. What a waste of time.

The bg fix does work.

fffe:
No, it's still leaving zombies.

```
PID   USER     TIME  COMMAND
[...]

PR #1849, opened 2026-06-26
fix background jobs

fffe:
execute.*.bg has been broken since the posix_spawn change, skip the immediate wait and add a reaper task to clean up later.

this solves my immediate problem of rtorrent deadlocking when running a completion job, but I'm not sure it's the right way to do it. Partially addresses #1848

### DIFF
--- a/src/rpc/exec_file.cc
+++ b/src/rpc/exec_file.cc
@@ -1,6 +1,8 @@
 #include "config.h"
 
+#include <cassert>
 #include <cerrno>
+#include <csignal>
 #include <cstring>
 #include <fcntl.h>
 #include <spawn.h>
@@ -19,6 +21,15 @@ extern char** environ;
 namespace rpc {
 
 // TODO: Access fd through torrent logging?
+
+void
+ExecFile::reap_background() {
+  while (true) {
+    pid_t pid = waitpid(-1, nullptr, WNOHANG);
+    if (pid == 0 || (pid == -1 && errno != EINTR))
+      break;
+  }
+}
 
 int
 ExecFile::execute(const char* file, char* const* argv, int flags) {
@@ -40,8 +51,13 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
 
   int pipe_fd[2];
 
-  if ((flags & flag_capture) && pipe(pipe_fd))
-    throw torrent::input_error("ExecFile::execute(...) Pipe creation failed.");
+  if (flags & flag_capture) {
+    if (flags & flag_background)
+      throw torrent::internal_error("ExecFile::execute(...) cannot capture a background task.");
+
+    if (pipe(pipe_fd) == -1)
+      throw torrent::input_error("ExecFile::execute(...) Pipe creation failed.");
+  }
 
   auto clean_fn = [pipe_fd, flags]() {
       if (flags & flag_capture) {
@@ -71,18 +87,20 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
     // Ensure the write end of the pipe is closed in the child after duplicating.
     posix_spawn_file_actions_addclose(&actions, pipe_fd[1]);
     posix_spawn_file_actions_addclose(&actions, pipe_fd[0]);
-
-  } else if (m_log_fd != -1) {
-    posix_spawn_file_actions_adddup2(&actions, m_log_fd, 1);
-
-  } else {
+  } else if (m_log_fd == -1) {
     posix_spawn_file_actions_addopen(&actions, 1, "/dev/null", O_RDWR, 0);
+  } else if (flags & flag_background) {
+    posix_spawn_file_actions_addopen(&actions, 1, "/dev/null", O_RDWR, 0);
+  } else {
+    posix_spawn_file_actions_adddup2(&actions, m_log_fd, 1);
   }
 
-  if (m_log_fd != -1) {
-    posix_spawn_file_actions_adddup2(&actions, m_log_fd, 2);
-  } else {
+  if (m_log_fd == -1) {
     posix_spawn_file_actions_addopen(&actions, 2, "/dev/null", O_RDWR, 0);
+  } else if (flags & flag_background) {
+    posix_spawn_file_actions_addopen(&actions, 2, "/dev/null", O_RDWR, 0);
+  } else {
+    posix_spawn_file_actions_adddup2(&actions, m_log_fd, 2);
   }
 
   posix_spawnattr_t attr;
@@ -92,9 +110,32 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
   // automatically. If you want to explicitly enforce a clean slate, modern systems support
   // POSIX_SPAWN_CLOEXEC_DEFAULT.
 
+  short spawn_flags = 0;
+
 #ifdef POSIX_SPAWN_CLOEXEC_DEFAULT
-  posix_spawnattr_setflags(&attr, POSIX_SPAWN_CLOEXEC_DEFAULT);
+  spawn_flags |= POSIX_SPAWN_CLOEXEC_DEFAULT;
 #endif
+
+  if (flags & flag_background) {
+    spawn_flags |= POSIX_SPAWN_SETPGROUP;
+    posix_spawnattr_setpgroup(&attr, 0);
+  }
+
+  posix_spawnattr_setflags(&attr, spawn_flags);
+
+  // Block SIGCHLD around the synchronous foreground wait so the SIGCHLD handler's waitpid(-1)
+  // can't reap (and discard the status of) the child we are about to wait on. Background tasks
+  // are left unblocked; the handler reaps them.
+  sigset_t old_mask;
+
+  if (!(flags & flag_background)) {
+    assert(torrent::this_thread::thread() == torrent::main_thread::thread());
+
+    sigset_t block_mask;
+    sigemptyset(&block_mask);
+    sigaddset(&block_mask, SIGCHLD);
+    pthread_sigmask(SIG_BLOCK, &block_mask, &old_mask);
+  }
 
   pid_t child_pid{};
 
@@ -104,8 +145,18 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
   posix_spawnattr_destroy(&attr);
 
   if (spawn_status != 0) {
+    if (!(flags & flag_background))
+      pthread_sigmask(SIG_SETMASK, &old_mask, nullptr);
+
     clean_fn();
     throw torrent::input_error("ExecFile::execute(...) posix_spawn failed: " + std::string(std::strerror(spawn_status)));
+  }
+
+  if (flags & flag_background) {
+    if (m_log_fd != -1)
+      result = write(m_log_fd, "\n--- Background task ---\n", sizeof("\n--- Background task ---\n"));
+
+    return 0;
   }
 
   if (flags & flag_capture) {
@@ -132,19 +183,29 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
   }
 
   int status;
+  int wait_errno = 0;
 
   while (waitpid(child_pid, &status, 0) == -1) {
-    switch (errno) {
-    case EINTR:
+    if (errno == EINTR)
       continue;
-    case ECHILD:
-      throw torrent::internal_error("ExecFile::execute(...) waitpid failed with ECHILD, child process not found.");
-    case EINVAL:
-      throw torrent::internal_error("ExecFile::execute(...) waitpid failed with EINVAL.");
-    default:
-      throw torrent::internal_error("ExecFile::execute(...) waitpid failed with unexpected error: " + std::string(std::strerror(errno)));
-    }
-  };
+
+    wait_errno = errno;
+    break;
+  }
+
+  if (!(flags & flag_background))
+    pthread_sigmask(SIG_SETMASK, &old_mask, nullptr);
+
+  switch (wait_errno) {
+  case 0:
+    break;
+  case ECHILD:
+    throw torrent::internal_error("ExecFile::execute(...) waitpid failed with ECHILD, child process not found.");
+  case EINVAL:
+    throw torrent::internal_error("ExecFile::execute(...) waitpid failed with EINVAL.");
+  default:
+    throw torrent::internal_error("ExecFile::execute(...) waitpid failed with unexpected error: " + std::string(std::strerror(wait_errno)));
+  }
 
   // Check return value?
   if (m_log_fd != -1) {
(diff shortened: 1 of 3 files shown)


src/rpc/exec_file.cc
```diff
@@ -38,6 +67,11 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
     result = write(m_log_fd, "\n---\n", sizeof("\n---\n"));
   }
 
+  if (flags & flag_background)
+    flags &= ~flag_capture;
```
rakshasa: This should throw internal errors.

src/rpc/exec_file.cc
```diff
@@ -38,6 +67,11 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
...
+
+  int child_log_fd = (flags & flag_background) ? -1 : m_log_fd;
```
rakshasa: This makes things harder to read, just add flag_background to the if/else blocks below.

src/rpc/exec_file.cc
```diff
@@ -108,6 +142,20 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
...
+  if (flags & flag_background) {
+    assert(torrent::this_thread::thread() == torrent::main_thread::thread());
+
+    if (m_log_fd != -1)
+      result = write(m_log_fd, "\n--- Background task ---\n", sizeof("\n--- Background task ---\n"));
+
+    m_background_pids.push_back(child_pid);
+
+    if (!m_background_task.is_scheduled())
+      torrent::this_thread::scheduler()->wait_for_ceil_seconds(&m_background_task, std::chrono::seconds(5));
```
rakshasa: With thousands of torrents, this could quickly cause a mess of zombie processes.

We need to ignore / handle SIGCHLD, and set:

```
posix_spawnattr_setflags(&attr, POSIX_SPAWN_SETPGROUP);
posix_spawnattr_setpgroup(&attr, 0); 
```

fffe:
Waiting for this to compile, is this how you were thinking?

src/rpc/exec_file.cc
```diff
@@ -20,6 +22,19 @@ namespace rpc {
...
+ExecFile::reap_background() {
+  for (auto itr = m_background_pids.begin(); itr != m_background_pids.end(); ) {
+    int   status;
+    pid_t result = waitpid(*itr, &status, WNOHANG);
+
+    if (result == 0 || (result == -1 && errno == EINTR))
+      itr++;
+    else
+      itr = m_background_pids.erase(itr);
+  }
```
rakshasa, line 31: This seems unnecessarily inefficient, we can use `-1` as the pid.

We could also use a lock-free list using std::atomic, or rather two lists.

ExecFile adds to one of them, the signal handler pops the pid it received from `waitpid()`.

To handle cases where signal handler gets called before ExecFile adds the pid entry, set an atomic flag before posix_spawn and clear it after adding pid to list.

The signal handler sets a different flag in the same atomic int to indicate it is processing SIGCHLD. If it can't find the pid in the spawn list, it adds the pid to an 'unknown pid' list.

ExecFile uses the atomic flag set by signal handler to wait for signal handler to finish, then checks the 'unknown pid' list.

If it finds its the new pid there, it removes it from both lists.

We would also need to change how waitpid is handled by non-background processes.

fffe:
> This seems unnecessarily inefficient, we can use -1 as the pid.

Yeah. I was trying to keep the change minimal, but I've modified it to use -1 for background tasks.

> We could also use a lock-free list using std::atomic, or rather two lists.

I think this over-complicates things... at least with the current core code, ExecFile::execute and the reaper both run on the main thread. I've tried to split the difference: blocking executions are handled mostly the same way, and the reaper is simplified.

Blocking SIGCHLD during sync execution avoids the only potential race (signal handler firing mid-synchronous-execute()). Because it's blocking not ignoring, any signals from background tasks exiting will land and trigger the reaper when the sync waitpid() returns and the handler is unblocked.

It also means that the reaper never triggers on sync executions and can just eat everything, which avoids tricky handler code... malloc isn't safe in a signal handler, synchronizing list access would suck.

src/rpc/exec_file.cc
```diff
@@ -20,6 +22,15 @@ namespace rpc {
 
 // TODO: Access fd through torrent logging?
 
+void
+ExecFile::reap_background() {
+  pid_t pid;
+
+  do {
+    pid = waitpid(-1, nullptr, WNOHANG);
+  } while (pid > 0 || (pid == -1 && errno == EINTR));
```
rakshasa: Use `if` within the while loop to make it easier to read.

Just make it a `while (true)` loop and set pid within it, it's easier to read.

src/rpc/exec_file.cc
```diff
@@ -40,8 +51,13 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
 
   int pipe_fd[2];
 
-  if ((flags & flag_capture) && pipe(pipe_fd))
-    throw torrent::input_error("ExecFile::execute(...) Pipe creation failed.");
+  if (flags & flag_capture) {
+    if (flags & flag_background)
+      throw torrent::internal_error("ExecFile::execute(...) cannot capture a background task.");
+
+    if (pipe(pipe_fd))
```
rakshasa: Do explicit checks of return value being `-1`.

src/rpc/exec_file.cc
```diff
@@ -71,15 +87,13 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
...
-  } else if (m_log_fd != -1) {
+  } else if (m_log_fd != -1 && !(flags & flag_background)) {
     posix_spawn_file_actions_adddup2(&actions, m_log_fd, 1);
-
   } else {
     posix_spawn_file_actions_addopen(&actions, 1, "/dev/null", O_RDWR, 0);
   }
 
-  if (m_log_fd != -1) {
+  if (m_log_fd != -1 && !(flags & flag_background)) {
```
rakshasa: I'd rather have the two if checks here and above be inverted and separated, so that it's easier to read even if it duplicates adddup.

src/rpc/exec_file.cc
```diff
@@ -92,10 +106,33 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
...
+    spawn_flags |= POSIX_SPAWN_SETPGROUP;
+    posix_spawnattr_setpgroup(&attr, 0);
+  }
+
+  posix_spawnattr_setflags(&attr, spawn_flags);
+
+  // Block SIGCHLD around the synchronous foreground wait so the SIGCHLD handler's waitpid(-1)
+  // can't reap (and discard the status of) the child we are about to wait on. Background tasks
+  // are left unblocked; the handler reaps them.
+  sigset_t old_mask;
```
rakshasa, line 129: This is the part I don't like, and why I want an atomic linked list.

Not just to catch any future potential bugs if someone adds forking elsewhere, but also because I'm not entirely sure if e.g. C++ async or other threads might potentially still listen to this signal.

 So I'd rather want a proper, atomic, way of dealing with pids. (While also catching unexpected pids)

src/rpc/exec_file.cc
```diff
@@ -104,10 +141,20 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
...
+    if (!(flags & flag_background))
+      pthread_sigmask(SIG_SETMASK, &old_mask, nullptr);
+
     clean_fn();
     throw torrent::input_error("ExecFile::execute(...) posix_spawn failed: " + std::string(std::strerror(spawn_status)));
   }
 
+  if (flags & flag_background) {
+    if (m_log_fd != -1)
+      result = write(m_log_fd, "\n--- Background task ---\n", sizeof("\n--- Background task ---\n"));
```
rakshasa, line 157: We should be able to use the log fd, even if it is a background process.

src/rpc/exec_file.cc
fffe: I don't think you can have that without making some other tradeoffs.

You're pretty limited in what you can do inside a signal handler, so either you're manipulating a fixed-length list of pids, or you're setting a flag and returning so something outside the handler can do something about it.

Neither of those really feels like an improvement, and that's a lot of complexity just in case someone adds another fork()

src/rpc/exec_file.cc
fffe: I don't follow, this is using the log fd?

src/rpc/exec_file.cc
fffe: Do you mean like this?

```
...
} else if (m_log_fd == -1) {
  posix_spawn_file_actions_addopen(&actions, 1, "/dev/null", O_RDWR, 0);
} else if (flags & flag_background) {
  posix_spawn_file_actions_addopen(&actions, 1, "/dev/null", O_RDWR, 0);
} else {
  posix_spawn_file_actions_adddup2(&actions, m_log_fd, 1);
}
```

Can do.

src/rpc/exec_file.cc
fffe: > I'm not entirely sure if e.g. C++ async or other threads might potentially still listen to this signal

Hmm... actually, I don't think anything could be, since 59fe935.

rakshasa:
```
#include <atomic>
#include <iostream>
#include <memory>
#include <thread>
#include <vector>

template <typename K, typename V>
class AtomicMapList {
private:
    struct Node {
        K key;
        V value;
        // Atomic shared_ptr manages the life of the next node safely
        std::atomic<std::shared_ptr<Node>> next{nullptr};

        Node(const K& k, const V& v) : key(k), value(v) {}
    };

    // Head of the list
    std::atomic<std::shared_ptr<Node>> head{nullptr};

public:
    AtomicMapList() = default;
    
    // Disable copying to protect atomic invariants
    AtomicMapList(const AtomicMapList&) = delete;
    AtomicMapList& operator=(const AtomicMapList&) = delete;

    // Insert at the front (Thread-Safe)
    void push_front(const K& key, const V& value) {
        auto newNode = std::make_shared<Node>(key, value);
        
        // Load current head into newNode->next
        newNode->next.store(head.load(std::memory_order_relaxed), std::memory_order_relaxed);

        // CAS loop to safely swap the new node into the head position
        while (!head.compare_exchange_weak(newNode->next, newNode,
                                           std::memory_order_release,
                                           std::memory_order_relaxed)) {
            // Loop automatically updates newNode->next on failure and retries
        }
    }

    // Find a key and return its value via a bool success flag (Thread-Safe)
    bool find(const K& key, V& out_value) const {
        // Increment reference count atomically to guarantee node stays alive
        std::shared_ptr<Node> current = head.load(std::memory_order_acquire);

        while (current != nullptr) {
            if (current->key == key) {
                out_value = current->value;
                return true;
            }
            current = current->next.load(std::memory_order_acquire);
        }
        return false;
    }

    // Erase a node by matching its key (Thread-Safe)
    bool erase(const K& key) {
        while (true) {
            std::shared_ptr<Node> current = head.load(std::memory_order_acquire);
            
            // Case 1: List is empty
            if (current == nullptr) return false;

            // Case 2: The head node matches the target key
            if (current->key == key) {
                std::shared_ptr<Node> nextNode = current->next.load(std::memory_order_acquire);
                // Attempt to advance head past the deleted node
                if (head.compare_exchange_weak(current, nextNode,
                                               std::memory_order_release,
                                               std::memory_order_relaxed)) {
                    return true; // Node successfully detached
                }
                continue; // CAS failed due to concurrent modification; retry outer loop
            }

            // Case 3: Key is deeper in the list
            std::shared_ptr<Node> pred = current;
            current = pred->next.load(std::memory_order_acquire);

            while (current != nullptr) {
                if (current->key == key) {
                    std::shared_ptr<Node> nextNode = current->next.load(std::memory_order_acquire);
                    
                    // Attempt to bypass 'current' node by pointing 'pred->next' directly to 'nextNode'
                    if (pred->next.compare_exchange_weak(current, nextNode,
                                                         std::memory_order_release,
                                                         std::memory_order_relaxed)) {
                        return true; // Successfully unlinked
                    }
                    break; // pred->next changed under our feet; break inner loop to restart from head
                }
                pred = current;
                current = current->next.load(std::memory_order_acquire);
            }
            
            // If we traversed the entire list without finding the key, it doesn't exist
            if (current == nullptr) return false;
        }
    }
};

int main() {
    AtomicMapList<int, std::string> map;

    // Populate data
    map.push_front(1, "One");
    map.push_front(2, "Two");
    map.push_front(3, "Three");

    // Concurrent read, erase, and write execution
    std::thread t1([&]() {
        std::string val;
        if (map.find(2, val)) std::cout << "Thread 1 Found: " << val << "\n";
    });

    std::thread t2([&]() {
        if (map.erase(2)) std::cout << "Thread 2 successfully erased Key 2\n";
    });

    std::thread t3([&]() {
        map.push_front(4, "Four");
    });

    t1.join();
    t2.join();
    t3.join();

    // Verify final state
    std::string finalVal;
    std::cout << "Key 2 exists? " << (map.find(2, finalVal) ? "Yes" : "No") << "\n";
    std::cout << "Key 4 exists? " << (map.find(4, finalVal) ? "Yes (" + finalVal + ")" : "No") << "\n";

    return 0;
}
```

Should work on C++20.

If we use std::async in main thread it will inherit the SIGCHLD handler.

fffe: There's an error in `push_front` but this isn't signal-safe.

`std::make_shared` allocates memory, `std::atomic<std::shared_ptr<Node>>` isn't lock-free (and triggers a free when it's eventually destroyed), etc.

There's not much that you can safely do from a signal handler.

rakshasa: You have main thread add space to the linked list of arrays of atomic pid keys, and if the signal handler gets SIGCHLD before main thread then there's an empty atomic pid for it to store it in.

There's no need to rely on the signal handler alloc/freeing memory. Just do it before posix_spawn and clean up later in main thread.

fffe: even if you're pre-allocating memory, most of this isn't on the signal-safe list (UB), and at least the shared_ptr args aren't lock-free (UB).

If you're using std::async on the main thread, you'd need to block the signals before spawning the thread anyway, like you did in https://github.com/rakshasa/rtorrent/commit/59fe93515ea7dece8a8b39d2f761624deada0cb

rakshasa: If `std::atomic<T>::is_always_lock_free` is true, it's signal-safe.

So `std::array<std::atomic<pid_t>>` is safe.

rakshasa:
`fix/exec-background` should fix the background spawning (which got dropped during the change)

I think I'll have to add a separate process for this, and use shm-ipc. There seems to be no clean way of handling background processes and normal waitpid, etc.

rakshasa:
It seems to work correctly when running `fix/exec-background`

fffe:
Yeah, that's a much better quick fix, ignoring matches the old behavior. Oof, I should have thought of that.

I'll close this out. Thanks for the feedback.

PR #1852, opened 2026-06-26
Add support for posix_spawn_file_actions_addclosefrom_np

fffe:
Add support for posix_spawn_file_actions_addclosefrom_np on systems that support it (glibc >= 2.34, freebsd >= 13.1) as a defense-in-depth against leaking file descriptors to child processes.

This pairs with, but does not require, https://github.com/rakshasa/libtorrent/pull/797

Related to https://github.com/rakshasa/rtorrent/issues/1848

before:
```
bash-5.3$ rtxmlrpc execute.capture '' ls /proc/self/fd | wc -l
330
```

after:
```
bash-5.3$ rtxmlrpc execute.capture '' ls /proc/self/fd | wc -l
5
```


src/rpc/exec_file.cc
```diff
@@ -88,12 +88,13 @@ ExecFile::execute(const char* file, char* const* argv, int flags) {
...
-
-#ifdef POSIX_SPAWN_CLOEXEC_DEFAULT
+  // Try to avoid leaking open fds to the spawned process. Prefer POSIX_SPAWN_CLOEXEC_DEFAULT
+  // (macOS-only) or posix_spawn_file_actions_addclosefrom_np (glibc >= 2.34, FreeBSD >= 13.1).
+  // Other platforms like musl libc, OpenBSD and NetBSD must rely on explicit O_CLOEXEC.
+#if defined(POSIX_SPAWN_CLOEXEC_DEFAULT)
   posix_spawnattr_setflags(&attr, POSIX_SPAWN_CLOEXEC_DEFAULT);
+#elif defined(HAVE_POSIX_SPAWN_FILE_ACTIONS_ADDCLOSEFROM_NP)
+  posix_spawn_file_actions_addclosefrom_np(&actions, 3);
 #endif
```
rakshasa, line 101: You need to call addclosefrom first, before any adddup2 or it won't work.

fffe: Oops, you're right.

This is working on my test client, I guess because it's only passing <= 3 fds anyway, but the order is wrong. That's a quick fix.

PR #1861, opened 2026-07-03
Add RISC-V cacheline fallbacks

Xeonacid:
The Linux cacheline probe first tries <linux/cache.h>, but that header is not part of the installed uapi headers on Arch Linux. When the compile probe fails, configure falls back to a host_cpu mapping and currently aborts for riscv64 with:

  Unrecognized CPU architecture (riscv64) on Linux fallback path.

Handle riscv* in both cacheline fallback maps and use a 64-byte cacheline. That matches the Linux RISC-V kernel default L1_CACHE_SHIFT value of 6 and the value reported on a riscv64 machine through getconf LEVEL1_DCACHE_LINESIZE and sysfs cache coherency_line_size.


rakshasa:
Match the whole `riscv64` string.

rakshasa:
Or rather, copy the updated common.m4 in libtorrent.

Xeonacid:
> Or rather, copy the updated common.m4 in libtorrent.

Updated with the libtorrent one, thanks!

PR #1866, opened 2026-07-05
Encryption configuration refactor

j-c-m:
companion to https://github.com/rakshasa/libtorrent/pull/808

Add user-friendly encryption mode configuration

Introduce mode presets (disabled, permissive, enabled, strict) and
granular protocol.encryption.* commands mapping directly to
EncryptionPolicy. Deprecates protocol.encryption.set flag list and the
encryption= redirect. Default to permissive. Update docs and tests.


src/command_network.cc
```diff
@@ -34,13 +34,14 @@
...
+  const uint32_t none = torrent::option_find_string(torrent::OPTION_ENCRYPTION, "none");
+  uint32_t options_mask = none;
 
   for (const auto& arg : args) {
     uint32_t opt = torrent::option_find_string(torrent::OPTION_ENCRYPTION, arg.as_string().c_str());
 
-    if (opt == torrent::runtime::NetworkConfig::encryption_none)
-      options_mask = torrent::runtime::NetworkConfig::encryption_none;
+    if (opt == none)
+      options_mask = none;
```
rakshasa: We can just decide on `0` (or default initialized) being none, and avoid extra code here.

PR #1868, opened 2026-07-05
Add a log method

fffe:
This is a nice-to-have for daemon mode (`print` goes nowhere, avoids spawning an echo or something), it adds a log method that gives access to the built-in logging.

```
method.insert = d.log_event, simple|private,\
    "log.log = info, (cat, (argument.0), \": \", (d.name), \" [\", (d.hash), \"]\")"
method.set_key = event.download.inserted_new, !log, "d.log_event_sized=added"
```

gives

```
1526374908 I added: linux.iso (  1.0 GB) [7D97E98F8AF710C7E7FE703ABC8F639E0EE507C4]
```

### DIFF
--- a/src/command_logging.cc
+++ b/src/command_logging.cc
@@ -1,6 +1,7 @@
 #include "config.h"
 
 #include <fcntl.h>
+#include <iterator>
 #include <stdio.h>
 #include <unistd.h>
 #include <torrent/data/chunk_utils.h>
@@ -15,6 +16,7 @@
 #include "core/download.h"
 #include "core/download_list.h"
 #include "core/manager.h"
+#include "rpc/parse.h"
 #include "rpc/parse_commands.h"
 
 torrent::Object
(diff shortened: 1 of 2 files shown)


rakshasa:
```
       "method.insert.c_simple = group.insert_persistent_view,"
       "((view.add,((argument.0)))),((view.persistent,((argument.0)))),((group.insert,((argument.0)),((argument.0))))\n"
```

Try to rewrite the example using the above style (from main.cc).

The command name needs improvement, and this could be optimized by having a version that takes the category index. And the user can convert the string to value when calling method insert.

rakshasa:
And since you already got a list, just make it to concatenate them directly, instead of having to rely on `cat`.

`log_group::internal_print` variant should be added that just does a simple string print.

fffe:
do you have a preference for a name?

"log" makes sense, and it's under the log family... more to the point, everything else I came up with sucked.

rakshasa:
```
log.push
log.print
```

rakshasa:
Commit messages need to be descriptive one-line messages or I'll have to squash-merge them.

Also, `log.log` being the same word twice looks bad, `log.print` makes much more sense.

fffe:
Yeah, sorry. Didn't mean to push that earlier.

I think 'log.print' has the changes you wanted now.

src/command_logging.cc
```diff
@@ -100,6 +121,16 @@ log_vmmap_dump(const std::string& str) {
...
 initialize_command_logging() {
+  // Resolve stable group names to the current internal indices when commands are parsed.
+  for (int group = 0; group < torrent::LOG_GROUP_MAX_SIZE; ++group) {
+    if (group == torrent::LOG_NON_CASCADING)
+      continue;
+
+    std::string command_name = "log.group." + std::string(torrent::option_to_c_str_or_throw(torrent::OPTION_LOG_GROUP, group));
+
+    CMD2_ANY(command_name, [group](auto, auto) { return int64_t(group); });
+  }
```
rakshasa: This is too many commands being added.

This should be in command_dynamic.cc, and called `enum.log_group`, take a string argument and return the enum value.

Making `apply_log_print` support both string and value as log category argument would make this more flexible.

fffe:
Pairs with https://github.com/rakshasa/libtorrent/pull/848

PR #1889, opened 2026-07-15
Restrict SCGI trust to local transports

acts-1631:
TCP SCGI clients are trusted when they omit UNTRUSTED_CONNECTION, allowing a reachable listener to invoke privileged RPC methods such as execute.raw.

An unauthenticated network peer can execute an attacker-selected program as the rTorrent user.

Derive trust from the listener transport: TCP connections are untrusted, Unix-domain sockets remain trusted, and request headers can only reduce trust. Document the TCP restriction.

### DIFF
--- a/doc/rtorrent.rc
+++ b/doc/rtorrent.rc
@@ -136,5 +136,6 @@
 # schedule can be used to set permissions on the unix socket.
 #
 #network.scgi.open_port = "127.0.0.1:5000"
+# TCP SCGI requests are untrusted. Use a Unix socket for trusted local control.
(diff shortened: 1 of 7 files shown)


rakshasa:
I'd rather remove the ability to use inet sockets for scgi than add this kind of workaround.

acts-1631:
> I'd rather remove the ability to use inet sockets for scgi than add this kind of workaround.

No objection from me.

rakshasa:
Is this actually required?

That is to say, should SCGI ports be AF_UNIX only.

PR #1893, opened 2026-07-18
Update default encryption protocol settings.

### LINKED ISSUE #1891: New encryption policy options reported as obsolete
Using rTorrent 0.16.18/0.16.18 packaged by Arch Linux.

On startup, rTorrent is always reporting that [the new encryption policy options](https://github.com/rakshasa/rtorrent/releases/tag/v0.16.18) are obsolete.

This occurs even for a completely clean installation with only the encryption options specified in the `.rtorrent.rc` file.

```
protocol.encryption.set = require
```

```
protocol.encryption.set = handshake_require, stream_require
```

Both of these configurations produce the same warning:

```
Obsolete encryption options used, use 'handshake_{deny,allow,prefer,require}, stream_{deny,allow,prefer,require}' instead.
```

Apologies if I'm overlooking something.

stephenmk:
The current default, hard-coded encryption protocol settings are obsolete. This PR updates the settings to use the new options.

https://github.com/rakshasa/rtorrent/wiki/Protocol-Encryption-Configuration

> Default is `allow`, which defaults to plaintext while allowing encrypted handshake and message stream

---

Verified that the correct encryption options are set by running:

```
command> print=$protocol.encryption=
```

Resolves #1891

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -301,7 +301,7 @@ main(int argc, char** argv) {
        "schedule = low_diskspace,5,60,((close_low_diskspace,500M))\n"
        "schedule = prune_file_status,3600,86400,((system.file_status_cache.prune))\n"
 
-       "protocol.encryption.set=allow_incoming,prefer_plaintext,enable_retry\n"
+       "protocol.encryption.set=allow\n"
 
        "ui.color.focus.set=reverse\n"
     );


rakshasa:
`fix/various-release-issues` fixes this by removing that line completely.

PR #1898, opened 2026-07-21
Support disabling upload traffic completely with '-1' throttle value.

### LINKED ISSUE #1885: Allow setting global upload speed limit to 0kb/s (to zero, not unlimited)
This "feature" is supported by Transmission and is the only thing keeps me away from swithing to rTorrent long-term.
Due to increasing broadband speeds, seeding criteria on many trackers are being continuously eased. Some still require seed time (instead of dl/ul ratio) to be met, the ones I'm using explicitly state that even if uploading 0 bytes, meeting the seed time requirements alone is sufficient to avoid HnR penalties.

On these sites I have been using 0kb/s Transmission limit without issues for years. Unfortunately, with rTorrent, the same cannot be achieved, since '0' means unlimited ('-1' or unset in Transmission). I don't have any preference on how this should be configured ('-1' to zero while keeping '0' as unlimited for backward compatibility or via a new setting), but I really would like this to be possible with rTorrent too.

szab100:
Fixes #1885

Very simple change, does exactly what Transmission when max upload speed is set to '0', effectively throttling all uploads to 0kb/s (resulting in peers eventually timing out).

### DIFF
--- a/doc/rtorrent.rc
+++ b/doc/rtorrent.rc
@@ -30,7 +30,7 @@
 #throttle.max_uploads.set = 15
 
 # Global upload and download rate in KiB.
-# "0" for unlimited.
+# "0" for unlimited, "-1" to disable all upload traffic.
 #
 #throttle.global_down.max_rate.set_kb = 0
 #throttle.global_up.max_rate.set_kb = 0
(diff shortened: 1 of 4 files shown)


rakshasa:
This doesn't really stop piece requests from being sent, and then stalling.

Also, it's basically breaking one of the core assumptions of the BT protocol. If someone wants to keep a patch set for this due to legal reasons (disabling piece upload in PeerConnection), I don't mind.

However I'm not really accepting it as a PR.

szab100:
@rakshasa OK, got it, thanks. If anyone needs this, can apply. I just like that in Transmission's '0' limit value actually means zero uploads (while still fetching content with maximum speeds).

However, PTAL at https://github.com/rakshasa/libtorrent/pull/837 since that is crucial to me and most probably others behind proxy (I use serjs/go-socks5-proxy using a gluetun container's network). Without the patch, announcing fail/stops after a few mins and never retried.

PR #1899, opened 2026-07-21
fix background execution

fffe:
Taking another stab at https://github.com/rakshasa/rtorrent/issues/1848

Dropping the double fork() means background processes don't get reparented to init, which means we need to do something about them.

My two cents, the best option would be adding process tracking to the event loop (pidfd_open/EVFILT_PROC), but pidfd_open doesn't exist before linux 5.3. I don't know if you have a support window you care about.

As before, this version adds a reaper thread to wait on background processes. This version switches a signal handler for sigwait(), which behaves a little more pleasantly. It does install a (dummy) signal handler so that macOS doesn't just drop the signals, but the handler is never called. This version also maintains a pid list and waits only for what it spawns.

It also fixes another bug, where children of rtorrent were inheriting its signal settings instead of the defaults.

### DIFF
--- a/src/control.cc
+++ b/src/control.cc
@@ -99,6 +99,7 @@ Control::cleanup() {
 
   m_ui->cleanup();
   m_core->cleanup();
+  rpc::execFile.cleanup();
 
   display::Canvas::erase_std();
   display::Canvas::refresh_std();
@@ -108,6 +109,7 @@ Control::cleanup() {
 
 void
 Control::cleanup_exception() {
+  rpc::execFile.cleanup();
   display::Canvas::cleanup();
 }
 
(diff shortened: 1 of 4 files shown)


rakshasa:
Feels like the wrong direction.

Shm-ipc seems to be ready, so it might be an idea to try to use that to spawn a separate exec process. However we still need to clean up those background processes even then.

https://github.com/rakshasa/shm-ipc

fffe:
If you're dedicating an entire process to it, you could set the handler to ignored and call it a day... or just waitpid(-1, ...). That would give back the old behavior.

If you want to wait only on specific pids, I think you're stuck with the same crappy options.

edit: If you do plan to go in another direction, you should still cherry pick the handler/mask reset stuff.

rakshasa:
Merged alternative implementation.

PR #1908, opened 2026-08-02
Fix zombie processes from background execute commands.

xirvik:
Since the posix_spawn conversion, background execute children are never waited on and SIGCHLD keeps its default disposition, so every execute.throw.bg / execute.nothrow.bg call leaves a permanent zombie. 

SignalHandler::set_sigchild_ignore() exists but is never called — and wiring it would break the foreground path: SA_NOCLDWAIT auto-reap races the blocking waitpid(), which then fails with ECHILD and loses the exit status.

This tracks background pids and reaps them with waitpid(WNOHANG) at the next execute call, keeping all spawning and waiting on the main thread.

Repro: execute.nothrow.bg={sleep,1} → permanent <defunct> child. Verified: zombies reaped on the next execute; still-running background children unaffected; foreground exit status, execute.throw and execute.capture behavior unchanged.

### DIFF
--- a/src/rpc/exec_file.cc
+++ b/src/rpc/exec_file.cc
@@ -1,5 +1,6 @@
 #include "config.h"
 
+#include <algorithm>
 #include <cassert>
 #include <cerrno>
 #include <cstring>
@@ -23,9 +24,18 @@ namespace rpc {
 
 // TODO: Access fd through torrent logging?
 
+void
+ExecFile::reap_background_processes() {
+  m_background_pids.erase(std::remove_if(m_background_pids.begin(), m_background_pids.end(),
+                                         [](pid_t pid) { return ::waitpid(pid, nullptr, WNOHANG) != 0; }),
+                          m_background_pids.end());
+}
+
 int
 ExecFile::execute(const char* file, char* const* argv, int flags) {
   assert(!((flags & flag_capture) && (flags & flag_background)));
+
+  reap_background_processes();
 
   // Write the executed command and its parameters to the log fd.
   [[maybe_unused]] int result;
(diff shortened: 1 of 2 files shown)


rakshasa:
https://github.com/rakshasa/rtorrent/pull/1899

There have been several attempts at fixing this, however they all have flaws.

Since the separate worker process using ipc is going to take a long while, I'll copy the FdCloseQueue pattern for this.

xirvik:
> #1899
> 
> There have been several attempts at fixing this, however they all have flaws.

What was wrong with the old double fork that  is a quite standard solution?

rakshasa:
https://github.com/rakshasa/rtorrent/pull/1909

Try the above.

And yes, it was one of the reasons we were seeing fd corruption in libcurl, as fork has a (very hard to trigger) race condition with epoll/kqueue fds.

xirvik:
> #1909
> 
> Try the above.

I think it's a bit worse TBH.  

It does wait_time = std::max(10s, wait_time * 2) after every sweep and only resets to 1 s when something is actually reaped — and close_pid() wakes the worker only if the queue was empty. 

So the first sweep therefore runs while the child is still alive, the next attempt is ≥10 s later, then 20 s, 40 s…  

In my tests I can see that  there's s a constantly refilling pile of transient zombies, even eventually the are all cleaned up.

Also, those two throws (internal_error) would crash the process.

rakshasa:
Those two internal_errors are never supposed to happen.

It was supposed to be std::min, not max when updating time_wait, and adding new wait pids didn't properly interrupt the sleep loop.

Fixed, so it should work properly now.

As for this PR it isn't going to work for batches of 1000's of calls, e.g. when starting rtorrent or scheduled tasks on all torrents, as it tries to reap every single call. This causes O(n*m) complexity.

PR #1910, opened 2026-08-03
Fix deprecated command redirects (rtorrent -D)

maksverver:
Before this, rtorrent failed to start with -D with errors like:

    Tried to redirect to a key that doesn't exist: 'network.port_range.set'

I updated the redirects for "port_random" and "proxy_address", but removed "port_range" since it's already defined unconditionally on line 347 above.


rakshasa:
Please force-push an update to this PR that has a more concise commit message and merges master branch.

maksverver:
> Please force-push an update to this PR that has a more concise commit message and merges master branch.

Done.

PR #1911, opened 2026-08-03
Add system.sockets.min_generic, .reserved and .available commands.

xirvik:
They report the allocation budget the socket manager enforces, so a client can show the limit and validate a change instead of discovering it from a fault.

Requires https://github.com/rakshasa/libtorrent/pull/865

### DIFF
--- a/src/command_local.cc
+++ b/src/command_local.cc
@@ -243,6 +243,8 @@ initialize_command_local() {
   CMD_ANY         ("system.sockets.max_size",         [](auto, auto)        { return torrent::runtime::socket_manager()->max_size(); });
   CMD_ANY_VALUE_V ("system.sockets.max_size.set",     [](auto, auto& value) { return torrent::runtime::socket_manager()->set_max_size_and_adjust(value); });
   CMD_ANY_V       ("system.sockets.adjust_alloc",     [](auto, auto)        { torrent::runtime::socket_manager()->adjust_allocation(); });
+  CMD_ANY         ("system.sockets.reserved_alloc",   [](auto, auto)        { return torrent::runtime::socket_manager()->reserved_allocation(); });
+  CMD_ANY         ("system.sockets.available_alloc",  [](auto, auto)        { return torrent::runtime::socket_manager()->available_allocation(); });
 
   for (uint32_t i = 0; i < torrent::runtime::SocketManager::category_count; ++i) {
     auto category      = static_cast<torrent::runtime::socket_manager_category_t>(i);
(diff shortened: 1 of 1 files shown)


src/command_local.cc
```diff
@@ -243,6 +243,9 @@ initialize_command_local() {
   CMD_ANY         ("system.sockets.max_size",         [](auto, auto)        { return torrent::runtime::socket_manager()->max_size(); });
   CMD_ANY_VALUE_V ("system.sockets.max_size.set",     [](auto, auto& value) { return torrent::runtime::socket_manager()->set_max_size_and_adjust(value); });
   CMD_ANY_V       ("system.sockets.adjust_alloc",     [](auto, auto)        { torrent::runtime::socket_manager()->adjust_allocation(); });
+  CMD_ANY         ("system.sockets.min_generic",      [](auto, auto)        { return torrent::runtime::socket_manager()->min_generic_allocation(); });
+  CMD_ANY         ("system.sockets.reserved",         [](auto, auto)        { return torrent::runtime::socket_manager()->reserved_allocation(); });
+  CMD_ANY         ("system.sockets.available",        [](auto, auto)        { return torrent::runtime::socket_manager()->available_allocation(); });
```
rakshasa: `system.sockets.generic.min_alloc` (or size? also rename to `generic_min_allocation()`)

`system.sockets.reserved_alloc` and `system.sockets.unallocated_alloc` or `free_alloc`?

`system.socket.*` namespace isn't necessarily going to just be for allocations

rakshasa (dismissed):

rakshasa:
Github is putting a rather negative color on a simple request for changes to a PR, not sure why it didn't just let me use the "comment" option.

src/command_local.cc
xirvik: @rakshasa I've renamed them. 

On on "available" though: What I need is the actual ceiling that ruTorrent can set - so it's not the same as free if there's some reservation that is hidden from clients, like this:

```
available_alloc = max_open - generic.min_alloc - reserved_alloc - 8
```

Ultimately what matter from the client point of view is the valid range for the set call, not internals.

Empirically tested:

| `ulimit -n` | `generic.min_alloc` | `reserved_alloc` | `available_alloc` | implied files ceiling | measured |
|---|---|---|---|---|---|
| 1024 | 512 | 128 | 376 | 280 | 280 accepted, 281 refused |
| 16384 | 12288 | 512 | 3576 | 3320 | 3320 accepted, 3321 refused |
| 32768 | 12288 | 512 | 19960 | 19704 | 19704 accepted, 19705 refused |

rakshasa:
Merged with fix.

PR #1916, opened 2026-08-07
Add string.* commands for text manipulation

### LINKED ISSUE #1366: Add built-in functions to work with strings
I've found at the [docs](https://rtorrent-docs.readthedocs.io/en/latest/cmd-ref.html#string-functions) that there are functions like `strings.startswith`, but they only supported in rtorrent-ps

It would be great to have such built-ins in original rtorrent codebase

Use case: I need to automate trackers list management. To be precise, I need to remove some trackers from torrent file on its insertion. Many of trackers in our country add the `retracker.local` tracker to the trackers list (it's something the local peer discovery and bep22 were created for at the same time). But nowadays this technique is no longer used by the ISPs. Hence, this trackers in torrents generates the errors and annoys me at the Flood UI with red rows

My solution is based on adding a handler to `event.download.inserted` event and then via `t.multicall` with conditional branch to disable that trackers

kunalkakade:
Closes #1366.

Adds a set of `string.*` commands for working with text from the configuration file. The names and signatures follow the ones documented for rTorrent-PS, so existing configs that rely on them can be used with upstream rTorrent unchanged.

## Commands

| Command | Signature | Result |
| --- | --- | --- |
| `string.length` | `«text»` | number of utf-8 characters |
| `string.equals` | `«text», «other»[, ...]` | `1` if any argument matches |
| `string.starts_with` | `«text», «prefix»[, ...]` | `1` if any prefix matches |
| `string.ends_with` | `«text», «tail»[, ...]` | `1` if any tail matches |
| `string.contains` | `«haystack», «needle»[, ...]` | `1` if any needle is found |
| `string.contains_i` | `«haystack», «needle»[, ...]` | as above, ascii case-insensitive |
| `string.substr` | `«text»[, «pos»[, «count»[, «default»]]]` | substring, negative `pos` counts from the end |
| `string.split` | `«text», «delim»` | list of parts |
| `string.join` | `«delim»[, «object»[, ...]]` | concatenation, lists are flattened |
| `string.lpad` / `string.rpad` | `«text», «length»[, «padding»]` | padded text |
| `string.strip` / `string.lstrip` / `string.rstrip` | `«text»[, «strippable»[, ...]]` | trimmed text |
| `string.map` | `«text», {«old», «new»}[, ...]` | replacement on a whole-string match |
| `string.replace` | `«text», {«old», «new»}[, ...]` | all occurrences replaced |

All of them are pure functions with no side effects, so they are registered as `mark_safe`.

Character counts and positions are in utf-8 characters rather than bytes, which matters for `string.length`, `string.substr`, `string.split` with an empty delimiter, the padding commands, and the strippable set.

`string.shorten` is deliberately left out: its default maximum length is not well defined and it is display-specific rather than text manipulation. It can follow separately if wanted.

## Layout

- `src/command_string.cc`: the implementation, registered from `initialize_commands()`.
- `test/src/test_command_string.{cc,h}`: 14 test cases in the `rtorrent_Test_Src` suite.
- `doc/manual/string_functions.md`: reference documentation with examples.

## Testing

`make check` passes. The new suite covers each command, its argument-count and type errors, utf-8 behaviour, and a `test_config_syntax` case that runs the commands through `rpc::parse_command_single` so the real configuration-file syntax is exercised, including the `{old, new}` blocks and nested commands.

Beyond the unit tests, the built binary was run against a torrent carrying an announce list with `retracker.local` in it, which is the use case from the issue:

```
method.set_key = event.download.inserted, drop_retracker, \
  ((t.multicall, default, "branch=(string.contains,(t.url),retracker.local),((t.disable))"))
```

Reading back `t.is_enabled` for each tracker afterwards gives `1 0 1`, so only the matching tracker is disabled.

### DIFF
--- a/doc/manual/string_functions.md
+++ b/doc/manual/string_functions.md
@@ -1,0 +1,139 @@
+# String functions
+
+The `string.*` commands manipulate and inspect text from within the
+configuration file.
+They are pure functions with no side effects, so they can be nested freely and
+are safe to call over RPC.
+
+Every argument is converted to its string representation before use, which means
+numbers can be passed where text is expected.
+Commands that count characters, such as `string.length` and `string.substr`, count
+utf-8 characters rather than bytes.
+
+## string.length
+
+    # string.length = «text»
+
+    string.length = "héllo"    # 5
+
+Returns the number of utf-8 characters in the text.
+
+## string.equals
+
+    # string.equals = «text», «other»[, ...]
+
+    string.equals = (d.name), "first.iso", "second.iso"
+
+Returns `1` if the first argument equals any of the following arguments,
+otherwise `0`.
+
+## string.starts_with, string.ends_with
+
+    # string.starts_with = «text», «prefix»[, ...]
+    # string.ends_with = «text», «tail»[, ...]
+
+    string.starts_with = (t.url), "http://", "https://"
+    string.ends_with = (d.name), ".iso"
+
+Returns `1` if the text begins, or ends, with any of the given prefixes or
+tails.
+
+## string.contains, string.contains_i
+
+    # string.contains = «haystack», «needle»[, ...]
+    # string.contains_i = «haystack», «needle»[, ...]
+
+    string.contains = (t.url), "retracker.local"
+    string.contains_i = (t.url), "RETRACKER.local"
+
+Returns `1` if the haystack contains any of the needles.
+The `_i` variant compares case-insensitively, and only handles ascii.
+
+## string.substr
+
+    # string.substr = «text»[, «position»[, «count»[, «default»]]]
+
+    string.substr = "abcdef", 2, 3       # "cde"
+    string.substr = "abcdef", -2         # "ef"
+    string.substr = "abcdef", 10, 1, "?" # "?"
+
+Extracts a part of the text, starting at `position` and spanning `count`
+characters.
+The position defaults to the start of the text, and the count to the rest of it.
+A negative position is relative to the end of the text.
+If the position falls outside the text, the default value is returned instead,
+which is the empty string unless given.
+
+## string.split
+
+    # string.split = «text», «delimiter»
+
+    string.split = "a.b.c", "."    # {"a", "b", "c"}
+    string.split = "abc", ""       # {"a", "b", "c"}
+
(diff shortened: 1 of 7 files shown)


doc/manual/string_functions.md
```diff
@@ -0,0 +1,139 @@
...
+Returns `1` if the first argument equals any of the following arguments,
+otherwise `0`.
+
+## string.startswith, string.endswith
+
+    # string.startswith = «text», «prefix»[, ...]
+    # string.endswith = «text», «tail»[, ...]
+
+    string.startswith = (t.url), "http://", "https://"
+    string.endswith = (d.name), ".iso"
```
rakshasa: starts_with and ends_with

doc/manual/string_functions.md
```diff
@@ -0,0 +1,139 @@
...
+Every argument is converted to its string representation before use, which means
+numbers can be passed where text is expected.
+Commands that count characters, such as `string.len` and `string.substr`, count
+utf-8 characters rather than bytes.
+
+## string.len
+
+    # string.len = «text»
+
+    string.len = "héllo"    # 5
```
rakshasa: length

doc/manual/string_functions.md
```diff
@@ -0,0 +1,139 @@
...
+
+Returns the number of utf-8 characters in the text.
+
+## string.equals
+
+    # string.equals = «text», «other»[, ...]
+
+    string.equals = (d.name), "first.iso", "second.iso"
```
rakshasa, line 25: never mind, equals is right for this

doc/manual/string_functions.md
kunalkakade: Renamed to `string.length`.

doc/manual/string_functions.md
kunalkakade: Renamed to `string.starts_with` and `string.ends_with`.

That leaves `string.substr`, `string.lpad`/`string.rpad` and `string.lstrip`/`string.rstrip` as the remaining abbreviated names. Say the word if you want those spelled out too and I will rename them in the same pass.

rakshasa: Those other names are good abbreviations.

PR #1918, opened 2026-08-07
Add realpath variants of the commands that return paths

### LINKED ISSUE #1670: Add command to get realpath of data and session files
To make passing paths to scripts safer, add `realpath` variants of commands that return paths.

While properly implemented scripts called by `execute` should always do independent sanity checks on paths passed as arguments, many don't so rtorrent should include this.

kunalkakade:
Closes #1670.

Adds a `.realpath` variant to each command that returns a path, so a path can be resolved before it is handed to an `execute` script rather than relying on the script to sanity check it.

| Command | Returns |
| --- | --- |
| `d.base_path.realpath.or_empty` / `.or_throw` | resolved `d.base_path` |
| `d.directory.realpath.or_empty` / `.or_throw` | resolved `d.directory` |
| `d.tied_to_file.realpath.or_empty` / `.or_throw` | resolved `d.tied_to_file` |
| `d.loaded_file.realpath.or_empty` / `.or_throw` | resolved `d.loaded_file` |
| `f.frozen_path.realpath.or_empty` / `.or_throw` | resolved `f.frozen_path` |
| `session.path.realpath.or_empty` / `.or_throw` | resolved `session.path` |
| `directory.default.realpath.or_empty` / `.or_throw` | resolved `directory.default` |

All of them go through a single `resolve_path()` helper added next to the existing `expand_path()` in `globals.cc`, so a leading `~` is expanded first and then `realpath(3)` collapses symlinks and any `.` or `..` components. Each variant is marked safe for RPC, matching the command it mirrors.

## Missing paths

`realpath(3)` needs the path to exist, which is often not the case here, so both behaviours are offered rather than one being picked for the caller.

`.or_empty` returns an empty string, which keeps `d.multicall` usable over a view holding both started and unstarted downloads, where throwing would abort the whole call over a single torrent with no data on disk yet. It also matches `d.base_path`, which already returns empty for an empty file list.

`.or_throw` raises `input_error` naming the path that failed, for the case where an unresolvable path means the command should not run at all.

## Two commands deliberately left out

- `d.directory_base` returns the same value as `d.directory` (both are `CMD2_ON_FL(root_dir)`), so a variant would just be a duplicate.
- `f.path` is relative to the download root, so resolving it on its own is not meaningful. `f.frozen_path` is the full path and covers the case.

Happy to add either if you would rather have them for symmetry.

## Testing

`make check` passes, on the default configure and on `--with-xmlrpc-tinyxml2`. The new `TestCommandPath` suite covers symlink resolution, paths below a symlinked directory, `.`/`..`/duplicate separator collapsing, tilde expansion, the empty and missing and dangling-symlink cases, and `directory.default.realpath` through the command layer.

The download and file targets need a live download, so those were checked against a running build instead, using a torrent whose data sits behind symlinks:

```
dl/single.bin -> /elsewhere/single.bin
dl/multi      -> /elsewhere/multi
```

After the hash check:

```
=== multi
  base_path    = .../dl/multi
  base_path.or_empty = .../elsewhere/multi
  directory.rp = .../elsewhere/multi
  files.rp     = .../elsewhere/multi/a.bin  .../elsewhere/multi/b.bin
=== single.bin
  base_path    = .../dl/single.bin
  base_path.rp = .../elsewhere/single.bin
  directory.rp = .../dl
```

`d.tied_to_file.realpath`, `d.loaded_file.realpath`, `session.path.realpath` and `directory.default.realpath` were checked the same way.

### DIFF
--- a/src/command_download.cc
+++ b/src/command_download.cc
@@ -655,6 +655,8 @@ initialize_command_download() {
   CMD2_DL("d.base_path.as_binary",            [](auto* download, auto) { return retrieve_d_base_path(download).object_as_binary(); });
   CMD2_DL("d.base_path.or_base64",            [](auto* download, auto) { return retrieve_d_base_path(download).object_utf8_or_base64(); });
   CMD2_DL("d.base_path.or_as_binary",         [](auto* download, auto) { return retrieve_d_base_path(download).object_utf8_or_as_binary(); });
+  CMD2_DL("d.base_path.realpath.or_empty",    [](auto* download, auto) { return resolve_path(retrieve_d_base_path(download).str()); });
+  CMD2_DL("d.base_path.realpath.or_throw",    [](auto* download, auto) { return resolve_path_or_throw(retrieve_d_base_path(download).str()); });
   CMD2_DL("d.base_filename",                  [](auto* download, auto) { return retrieve_d_base_filename(download).str(); });
   CMD2_DL("d.base_filename.hex",              [](auto* download, auto) { return retrieve_d_base_filename(download).object_hex(); });
   CMD2_DL("d.base_filename.base64",           [](auto* download, auto) { return retrieve_d_base_filename(download).object_base64(); });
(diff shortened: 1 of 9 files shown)


src/command_download.cc
```diff
@@ -655,6 +655,7 @@ initialize_command_download() {
   CMD2_DL("d.base_path.as_binary",            [](auto* download, auto) { return retrieve_d_base_path(download).object_as_binary(); });
   CMD2_DL("d.base_path.or_base64",            [](auto* download, auto) { return retrieve_d_base_path(download).object_utf8_or_base64(); });
   CMD2_DL("d.base_path.or_as_binary",         [](auto* download, auto) { return retrieve_d_base_path(download).object_utf8_or_as_binary(); });
+  CMD2_DL("d.base_path.realpath",             [](auto* download, auto) { return resolve_path(retrieve_d_base_path(download).str()); });
```
rakshasa: These would be `.realpath.or_empty` (and perhaps we need `.realpath.or_throw` also)

kunalkakade: Split into `.realpath.or_empty` and `.realpath.or_throw` for all seven, so there is no bare `.realpath` left. `or_throw` raises `input_error` naming the path that failed.

Both are covered by the unit tests, and both were checked against a running build with symlinked torrent data.

rakshasa:
Could you force-push with merge conflict resolved? Or i can squash merge.

kunalkakade:
Force-pushed, rebased onto master with the conflict resolved. The only conflict was the two new entries in the \`rtorrent_Test_Src_SOURCES\` block of \`test/Makefile.am\`, now listing both \`test_command_path\` and \`test_command_string\`.

Branch is two commits on top of \`cfb9e08\` and the tree is identical to the merge you pushed. \`make check\` passes on the default configure and on \`--with-xmlrpc-tinyxml2\`, 29 tests including both suites.

PR #1928, opened 2026-08-10
Do not dereference missing text nodes in the tinyxml2 backend.

xirvik:
TL;DR Empty elements and element children crashed the parser on a null.

**Repro**

  Against master `224f06f` built with the tinyxml2 backend, each on a fresh daemon:

  ```
  <?xml version="1.0"?><methodCall><methodName/><params></params></methodCall>
  <?xml version="1.0"?><methodCall><methodName>system.client_version</methodName>
    <params><param><value><i8><x/></i8></value></param></params></methodCall>
  ```

 The XMLRPC parser reaches for a node's text without checking that the text node
  exists. Seven request shapes crash the daemon before the trust check runs:

  | payload | result |
  |---|---|
  | `<methodName/>` | SIGSEGV — `GetText()` is null and the result is compared against a `std::string`, so `strlen(NULL)` |
  | multicall entry whose methodName is `<string/>` | `std::logic_error: basic_string: construction from null is not valid` |
  | struct member with `<name/>` | same, the null becomes a map key |
  | `<i8><x/></i8>` | SIGSEGV |
  | `<boolean><x/></boolean>` | SIGSEGV |
  | `<string><x/></string>` | SIGSEGV |
  | `<base64><x/></base64>` | SIGSEGV |

  The existing guards check the wrong thing. `element_to_int` verifies
  `elem->FirstChild() != nullptr`, but an *element* child satisfies that and then
  `ToText()` returns null on the next line. The same shape appears in the string,
  boolean and base64 branches.

  This adds an `element_text_value()` helper that throws `XMLRPC_TYPE_ERROR` when a
  node carries no text, and uses it at those four sites. The two `GetText()` results
  that feed a `std::string` get explicit null checks: an empty `methodName` is a
  parse error, and an empty struct member name becomes an empty key rather than a
  null.

### DIFF
--- a/src/rpc/xmlrpc_tinyxml2.cc
+++ b/src/rpc/xmlrpc_tinyxml2.cc
@@ -49,13 +49,23 @@ element_access(const tinyxml2::XMLElement* elem, std::initializer_list<std::stri
   return result;
 }
 
+const char*
+element_text_value(const tinyxml2::XMLNode* node) {
+  auto text = node->ToText();
+
+  if (text == nullptr)
+    throw rpc_error(XMLRPC_TYPE_ERROR, "expected a text value");
+
+  return text->Value();
+}
+
 long long
 element_to_int(const tinyxml2::XMLNode* elem) {
   char* pos;
   if (elem->FirstChild() == nullptr) {
     throw rpc_error(XMLRPC_TYPE_ERROR, "unable to parse empty integer");
   }
-  auto str    = elem->FirstChild()->ToText()->Value();
+  auto str    = element_text_value(elem->FirstChild());
   auto result = std::strtoll(str, &pos, 10);
   if (pos == str || *pos != '\0')
     throw rpc_error(XMLRPC_TYPE_ERROR, "unable to parse integer value");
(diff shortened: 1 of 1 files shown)


src/rpc/xmlrpc_tinyxml2.cc
```diff
@@ -129,7 +139,9 @@ xml_value_to_object(const tinyxml2::XMLNode* elem) {
       if (name_element == nullptr)
         throw rpc_error(XMLRPC_PARSE_ERROR, "struct member missing name element");
 
-      map[name_element->GetText()] = std::move(xml_value_to_object(child->FirstChildElement("value")));
+      auto name_text = name_element->GetText();
+
+      map[name_text != nullptr ? name_text : ""] = std::move(xml_value_to_object(child->FirstChildElement("value")));
```
rakshasa: Seems nullptr should throw.

xirvik: > Seems nullptr should throw.

Ah yes, thanks, fixed

PR #1933, opened 2026-08-10
Refuse a multicall whose commands change the download list.

xirvik:
TL;DR Erasing a download left the loop dispatching on freed pointers.

  `d_multicall` copies the view into a `std::vector<core::Download*>` and then runs arbitrary command strings against each raw pointer. `DownloadList::erase` deletes  the download immediately, so a command that erases leaves the rest of the loop  dispatching on freed memory:

  ```
  ERROR: AddressSanitizer: heap-use-after-free  READ of size 8
    #0 torrent::Download::info() const            torrent/download.cc:33
    #2 initialize_command_download()::<lambda>    command_download.cc:668
  freed by thread T0 here:
    #1 core::DownloadList::erase(...)             core/download_list.cc:212
    #2 core::DownloadList::erase_ptr(...)         core/download_list.cc:190
  ```

  `d.multicall2("", "main", "d.erase=", "d.name=")` is enough — the erase and the
  read are in the same row. `d.multicall.filtered` has the same shape.

  `DownloadList` now carries a counter bumped whenever a download is inserted or removed, and both multicalls check it before each command, refusing with  `input_error` if the list moved. `clear()` needs its own bump because it does   `pop_back` and `delete` directly rather than going through `erase`.

**Note: Not totally sure of the implications - clearly the bug exists but is this the right solution?**

### DIFF
--- a/src/command_events.cc
+++ b/src/command_events.cc
@@ -242,10 +242,16 @@ d_multicall(const torrent::Object::list_type& args) {
   torrent::Object             resultRaw = torrent::Object::create_list();
   torrent::Object::list_type& result = resultRaw.as_list();
 
+  auto* download_list  = control->core()->download_list();
+  auto  change_counter = download_list->change_counter();
+
   for (auto download : dlist) {
     torrent::Object::list_type& row = result.insert(result.end(), torrent::Object::create_list())->as_list();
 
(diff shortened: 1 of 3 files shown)


rakshasa:
Seems core::DownloadList should hold shared_ptrs, and you either use weak_ptrs here, or check if ref count == 1 to see if it is deleted. The latter would be faster.

xirvik:
Abandoning in favor of 1937

PR #1934, opened 2026-08-10
Add network.local_port commands for tracker port reporting

Auska:
Mirror the network.local_address.* command family for ports:
- network.local_port            report the best-match local port
- network.local_port.set        set both the ipv4 and ipv6 local ports
- network.local_port.ipv4(.set) get/set the ipv4 local port only
- network.local_port.ipv6(.set) get/set the ipv6 local port only

Values are restricted to 0-65535; 0 (default) means unset, and trackers then report the listening port.

Why: behind multiple layers of NAT the port peers must connect to can differ from the port rtorrent is listening on. These commands let the user manually advertise the forwarded public port per address family, e.g.:
    network.local_port.ipv4.set = 6881
    network.local_port.ipv6.set = 6882

network.local_port is marked safe for RPC use alongside network.local_address.

### DIFF
--- a/src/command_network.cc
+++ b/src/command_network.cc
@@ -54,6 +54,14 @@ set_listen_port_range(const std::string& arg) {
   torrent::runtime::client_config()->set_listen_port_range(port_first, port_last);
 }
 
+uint16_t
+checked_local_port_value(int64_t value, const char* label) {
+  if (value < 0 || value > 65535)
+    throw torrent::input_error(std::string("Invalid ") + label + " port number.");
+
+  return static_cast<uint16_t>(value);
+}
+
 torrent::Object
 get_encryption() {
   auto encryption_modes = torrent::runtime::network_config()->encryption_modes();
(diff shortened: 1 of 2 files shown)


src/command_network.cc
```diff
@@ -437,6 +456,7 @@ initialize_command_network() {
   rpc::rpc.mark_safe("network.receive_buffer.size");
   rpc::rpc.mark_safe("network.bind_address");
   rpc::rpc.mark_safe("network.local_address");
+  rpc::rpc.mark_safe("network.local_port");
   rpc::rpc.mark_safe("network.xmlrpc.size_limit");
```
rakshasa, line 458: This is missing ipv4/ipv6 versions.

src/command_network.cc
```diff
@@ -379,6 +379,25 @@ initialize_command_network() {
...
+    if (value < 0 || value > 65535)
+      throw torrent::input_error("Invalid local port number.");
+    return nw_config->set_local_inet_port(value);
+  });
+  CMD_ANY         ("network.local_port.ipv6",                [nw_config](auto, auto)         { return nw_config->local_inet6_port(); });
+  CMD_ANY_VALUE_V ("network.local_port.ipv6.set",            [nw_config](auto, auto& value) {
+    if (value < 0 || value > 65535)
+      throw torrent::input_error("Invalid local port number.");
+    return nw_config->set_local_inet6_port(value);
+  });
```
rakshasa: The lambdas here should be a single line, e.g. use a helper function for validating port numbers. (pass char* string for "local", "local ipv4", etc for the error string.

PR #1937, opened 2026-08-14
Let the download list own its downloads through a shared pointer.

xirvik:
`d.multicall` copies the view into raw `Download*` and then runs arbitrary  commands against them, so a command that erases leaves the rest of the call dispatching on freed memory — `d.multicall2 "" main "d.erase=" "d.name="` is  enough to hit it.

  `DownloadList` now holds `shared_ptr<Download>`, and `Download` derives from`enable_shared_from_this` so the multicall can re-own the pointers it takes from  a view without searching the list. It checks `use_count() == 1` before each  command and skips that download once the list has dropped its reference. The  count only detects the erase, it does not keep the download usable.

 Supersedes #1933, which refused the whole multicall instead of skipping the erased entry.

Tested with both crash shapes, `make check`, and on a live instance: a two-row call with `d.name=` `d.erase=` `d.name=` returns the leading read, the erase result, no trailing read, and continues to the next row.

My personal rtorrent instance has been running stable with this for a few hours.

### DIFF
--- a/src/command_events.cc
+++ b/src/command_events.cc
@@ -51,7 +51,7 @@ apply_on_ratio(const torrent::Object& rawArgs) {
         !(max_ratio > 0 && total_upload * 100 > total_done * max_ratio))
       continue;
 
-    downloads.push_back(*itr);
+    downloads.push_back(itr->get());
   }
 
   auto ratio_command = "group." + group_name + ".ratio.command";
@@ -173,7 +173,7 @@ apply_close_low_diskspace(int64_t arg, uint32_t skip_priority) {
 
   torrent::FileList::cache_list cache;
 
-  for (auto download : *control->core()->download_list()) {
+  for (const auto& download : *control->core()->download_list()) {
     if (!download->is_downloading())
       continue;
     if (download->priority() >= skip_priority)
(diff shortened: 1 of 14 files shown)


src/core/download.h
```diff
@@ -13,7 +14,7 @@
 
 namespace core {
 
-class Download {
+class Download : public std::enable_shared_from_this<Download> {
```
rakshasa: Don't use `enable_shared_from_this` as it is a bad pattern.

xirvik: > Don't use `enable_shared_from_this` as it is a bad pattern.

Changed. Better?

src/command_events.cc
```diff
@@ -111,11 +117,11 @@ torrent::Object
 apply_remove_untied() {
   for (auto itr = control->core()->download_list()->begin(); itr != control->core()->download_list()->end(); ) {
     torrent::utils::FileStat fs;
-    const std::string& tied_to_file = rpc::call_command_string("d.tied_to_file", rpc::make_target(*itr));
+    const std::string& tied_to_file = rpc::call_command_string("d.tied_to_file", rpc::make_target(itr->get()));
 
     if (!tied_to_file.empty() && !fs.update(expand_path(tied_to_file))) {
       // Need to clear tied_to_file so it doesn't try to delete it.
-      rpc::call_command("d.tied_to_file.set", std::string(), rpc::make_target(*itr));
+      rpc::call_command("d.tied_to_file.set", std::string(), rpc::make_target(itr->get()));
```
rakshasa: You can simplify this PR a lot by just adding another `rpc::make_target()` overload.

src/ui/download_list.cc
```diff
@@ -145,7 +145,7 @@ DownloadList::activate_display(Display displayType) {
       return;
 
     } else {
-      Download* download = new Download(*current_view()->focus());
+      Download* download = new Download(current_view()->focus()->get());
```
rakshasa: This is a very bad pattern and names that can be confused, it should take the shared_ptr reference and get the pointer in the ctor.

src/rpc/command_map.h
```diff
@@ -104,6 +105,11 @@ inline target_type make_target(T target) {
   return target_type((int)target_type_id<T>::value, target);
 }
 
+template <typename T>
+inline target_type make_target(const std::shared_ptr<T>& target) {
+  return make_target(target.get());
+}
```
rakshasa, line 110: Change this to be explicitly Download as the type.

src/command_events.cc
```diff
@@ -138,9 +138,6 @@ apply_schedule(const torrent::Object::list_type& args, bool if_absent) {
   auto& arg2 = (itr++)->as_string();
   auto& arg3 = (itr++)->as_string();
 
-  if (if_absent && control->command_scheduler()->find(arg1) != control->command_scheduler()->end())
-    return torrent::Object();
```
rakshasa, line 142: Looks like a merge issue here, this shouldn't be removed.

PR #1940, opened 2026-08-15
Add schedule.if_absent for clients that re-register periodic tasks.

xirvik:
Context: https://github.com/Novik/ruTorrent/pull/3163

TL;DR That PR works around the problem, sort of, but the best solution is in rtorrent - just provide a way to setup a schedule if not present already, otherwise skip.

A client that registers its periodic tasks whenever it connects has no way to avoid restarting them. schedule replaces an entry that already exists and arms the new item at now plus the absolute argument, and there is no command to ask whether a key is already scheduled, so the countdown starts over on every connect. A web UI that registers an hourly task on each page load and is reloaded more often than that never reaches the first call.

schedule.if_absent takes the same arguments and leaves an existing entry alone, so re-registering is a no-op and the task keeps its original timing. Changing a schedule still goes through schedule, or schedule.remove first.

### DIFF
--- a/src/command_events.cc
+++ b/src/command_events.cc
@@ -128,7 +128,7 @@ apply_remove_untied() {
 }
 
 torrent::Object
-apply_schedule(const torrent::Object::list_type& args) {
+apply_schedule(const torrent::Object::list_type& args, bool if_absent) {
   if (args.size() != 4)
     throw torrent::input_error("Wrong number of arguments.");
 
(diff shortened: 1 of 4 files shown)


src/rpc/command_scheduler.cc
```diff
@@ -102,6 +102,19 @@ CommandScheduler::parse(const std::string& key,
...
+                                  const std::string& bufAbsolute,
+                                  const std::string& bufInterval,
+                                  const torrent::Object& command) {
+  if (find(key) != end())
+    return false;
+
+  parse(key, bufAbsolute, bufInterval, command);
+
+  return true;
+}
```
rakshasa: This code adds unnecessary bloat, just do the check in the caller.

PR #1950, opened 2026-08-18
Run load command lists under the trust state of the submitting connection

xirvik:
process_untrusted() lowers the trusted flag, runs the request and restores it on one call stack, but DownloadFactory does not load on that stack: it defers to the  scheduler and returns. By the time receive_success() runs the command list attached to the load, the request is long finished and the flag is back to  trusted, so those commands run trusted however the request arrived.
 
DownloadFactory now records rpc.is_trusted() when it is constructed and re-enters  that state around the command loop through an RAII scope, which also restores on  the throwing paths out of receive_success(). Anything not created during an RPC  request - watch directories, session loads, .rtorrent.rc - is constructed while  trust is true and keeps running trusted. process_untrusted() uses the same scope  instead of its own save/restore.

Not currently reachable: load.* is not in the safe set, so this is latent until it is marked, which is the point of fixing it first. Verified by temporarily  marking load.raw_start safe and sending it untrusted with execute.capture=/bin/touch,<marker> attached: before, the marker appears; after,  the load still succeeds and the command is refused. Watch-dir and rc-driven loads still run their command lists, and load.* remains refused untrusted.

### DIFF
--- a/src/core/download_factory.cc
+++ b/src/core/download_factory.cc
@@ -24,6 +24,7 @@
 #include "core/http_queue.h"
 #include "core/manager.h"
 #include "rpc/parse_commands.h"
+#include "rpc/rpc_manager.h"
 
 namespace core {
 
@@ -62,7 +63,8 @@ is_magnet_uri(const std::string& uri) {
 }
 
 DownloadFactory::DownloadFactory(Manager* m) :
-    m_manager(m) {
+    m_manager(m),
+    m_trusted(rpc::rpc.is_trusted()) {
 
   m_task_load.slot() = std::bind(&DownloadFactory::receive_load, this);
   m_task_commit.slot() = std::bind(&DownloadFactory::receive_commit, this);
@@ -325,8 +327,12 @@ DownloadFactory::receive_success() {
     if (torrent::log_groups[torrent::LOG_TORRENT_DEBUG].valid())
       log_created(download, rtorrent);
 
-    for (const auto& command : m_commands)
-      rpc::parse_command_multiple_std(command, rpc::make_target(download));
+    {
+      rpc::trust_scope scope(m_trusted);
+
+      for (const auto& command : m_commands)
+        rpc::parse_command_multiple_std(command, rpc::make_target(download));
+    }
 
     if (m_manager->download_list()->find(infohash) == m_manager->download_list()->end())
       throw torrent::input_error("The newly created download was removed.");
--- a/src/core/download_factory.h
+++ b/src/core/download_factory.h
@@ -75,6 +75,7 @@ private:
   bool                m_printLog{true};
   bool                m_isFile{};
   bool                m_initLoad{};
+  bool                m_trusted{true};
 
   command_list_type         m_commands;
   torrent::Object::map_type m_variables;
--- a/src/rpc/rpc_manager.cc
+++ b/src/rpc/rpc_manager.cc
@@ -135,17 +135,18 @@ RpcManager::process(RPCType type, const char* in_buffer, uint32_t length, slot_r
 
 bool
 RpcManager::process_untrusted(RPCType type, const char* in_buffer, uint32_t length, slot_response_callback callback) {
-  bool previous = m_trusted;
-  m_trusted = false;
+  trust_scope scope(false);
 
-  try {
-    bool result = process(type, in_buffer, length, callback);
-    m_trusted = previous;
-    return result;
-  } catch (...) {
-    m_trusted = previous;
-    throw;
-  }
+  return process(type, in_buffer, length, callback);
+}
+
+trust_scope::trust_scope(bool state) :
+    m_previous(rpc.is_trusted()) {
+  rpc.set_trusted(state);
+}
+
+trust_scope::~trust_scope() {
+  rpc.set_trusted(m_previous);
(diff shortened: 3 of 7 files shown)


rakshasa:
Seems to me that loading a torrent is an a trusted action, aren't you shifting the meaning of trusted / untrusted connections?

xirvik:
> Seems to me that loading a torrent is an a trusted action, aren't you shifting the meaning of trusted / untrusted connections?

  1. client sends load.raw_start with UNTRUSTED_CONNECTION: 1, attaching
     execute.capture=/bin/touch,/tmp/marker
  2. process_untrusted() lowers the flag, apply_load() queues the factory, returns,
     flag restored to trusted
  3. scheduler runs receive_success() later and executes the attached command list
  4. marker exists: execute.capture ran on behalf of an untrusted connection

  Measured with a test-only mark_safe("load.raw_start") that is not in this PR:

  - master: marker created
  - this PR: load succeeds, attached command refused
  - watch dir, session load, .rtorrent.rc: command lists still run

  This PR adds nothing to the safe set. With load.* unmarked it is a no-op.

  Workflow we want, and why this has to land first:

  - mobile client on an untrusted connection adds a torrent:
    load.raw_start with d.directory.set attached
  - needs load.raw_start and d.directory.set marked safe; neither is today
  - marking load.raw_start safe without this PR: every attached command runs
    trusted, including execute.capture, method.insert, import
  - with this PR: each attached command is checked against the safe set, so the
    safe set stays the only thing that decides what an untrusted client can do

rakshasa:
This is the wrong direction, instead of securing current load commands you should create a new load command for untrusted connections.

It should strip (or rather throw on) any non-standard BT keys and have a separate SCGI header enabling it.

xirvik:
> This is the wrong direction, instead of securing current load commands you should create a new load command for untrusted connections.
> 
That wouldn't help with the existing clients though. ruTorrent is no problem because I maintain it, but I can't do anything about the rest. So the RPC2 endpoint needs to be open for all those clients that depend on it, but in a way that doesn't just give execute permissions to anyone who can reach it...

rakshasa:
Separating things into trusted and untrusted connections means rtorrent needs to create certain boundaries for the latter.

Allowing loading of torrents using the old code paths greatly expands the attack surface, which breaks the promise of the "untrusted connection" feature.

So the focus isn't really on compatibility.

xirvik:
> 
> So the focus isn't really on compatibility.

It's necessary unless a safe client with no users is a valid outcome.  

Of course that compatibility leads some ugly compromises at least temporarily.

rakshasa:
There is no "safe client", only "safer client", and allowing load commands makes things way less safe.

That's why I'm insisting that a load command needs to be heavily restricted.

If you need to run commands for torrents loaded by untrusted connections then those commands should be pre-configured commands.

xirvik:
> That's why I'm insisting that a load command needs to be heavily restricted.

We agree on the goal clearly but not on how to get there. 

IMHO compatibility is crucial - otherwise users will keep using whatever rtorrent version lets them use their existing clients.

> If you need to run commands for torrents loaded by untrusted connections then those commands should be pre-configured commands.

Do you have a design in mind?

xirvik:
Closing since this doesn't seem to go in a good direction.

rakshasa:
Add a spesial safe load commands, disabled by default, can't run load commands straight like that in case they are (intentionally) malformed it throws and aborts partway so need to catch and remove torrent.

Possibly add a event.download.insert_safe (and _unsafe when not) that is run on it.

rakshasa:
The way `rpc::rpc.is_trusted()` is called is not good, it should be a different DownloadFactory function to load safe torrents.

rakshasa:
See `feature/untrusted-load`

xirvik:
> See `feature/untrusted-load`

I think the meaning of trusted here is inverted...  I mean what is trusted or not  is not the command, but rather the caller (the connection), so IMHO the commands shouldn't use trusted or untrusted, but maybe safe and unsafe (and the current set is of course unsafe).

rakshasa:
The trusted/untrusted is about the permissions the caller gets to have, untrusted means the download factory can only call safe commands.

It is allowed to load a torrent, but not trusted enough to call unsafe commands and various other things.

xirvik:
> The trusted/untrusted is about the permissions the caller gets to have, untrusted means the download factory can only call safe commands.

Yes; IMHO confusing naming if trusted/untrusted appears in the method names and that suggests it's the method that is trusted or not.

> It is allowed to load a torrent, but not trusted enough to call unsafe commands and various other things.

BTW about this: Setting per torrent directories (d.directory.set)  and labels ( d.custom1.set) are both necessary for any client and problematic (i.e. sanitization needed).

rakshasa:
> BTW about this: Setting per torrent directories (d.directory.set) and labels ( d.custom1.set) are both necessary for any client and problematic (i.e. sanitization needed).

Seems like we should have a whitelist of directories and a `d.directory.set_safe`. 

Mark the rpc command safe at the same time we lock the whitelist, and the whitelist cannot be changed. This means the user can't sneak in extra whitelisted directories in imported config files if it gets locked early.

It should have an option to apply to the normal set directory command.

PR #1951, opened 2026-08-18
Add network.scgi.trusted for the default trust of SCGI connections

xirvik:
TL;DR - the current system, as is, is exploitable. ruTorrent's httprpc proxy sanitizes and that seems enough for clients that use that proxy, but transdroid and many others demand access to the RPC2 endpoint. I'm doing small steps towards locking down RPC2 without breaking clients. 

The untrusted-connection flag added in 5989149 can currently only be asserted by the client, via the UNTRUSTED_CONNECTION header. An operator has no way to state a posture: a connection starts trusted, and a client that sends nothing gets full access to every command. This adds the operator side of that decision.

 network.scgi.trusted.set = no makes SCGI connections start untrusted, so a reverse proxy that forgets the header — or a listener exposed by accident — no longer yields a trusted RPC endpoint. The default is true, so nothing changes unless it is configured.

 The header semantics change with it: UNTRUSTED_CONNECTION=1 still always lowers trust, but =0 now restores only as far as the configured default rather than unconditionally setting trusted. Without that, a client sends 0 and the setting has no effect.

network.scgi.trusted and .set are deliberately not marked untrusted-safe, so a connection cannot read or change its own trust level.

### DIFF
--- a/src/command_network.cc
+++ b/src/command_network.cc
@@ -413,6 +413,9 @@ initialize_command_network() {
   CMD_ANY         ("network.scgi.gzip.min_size",             [](auto, auto)                  { return rpc::rpc.scgi_min_compress_size(); });
   CMD_ANY_VALUE_V ("network.scgi.gzip.min_size.set",         [](auto, auto& arg)             { return rpc::rpc.set_scgi_min_compress_size(arg); });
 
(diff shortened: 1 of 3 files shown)


xirvik:
Closing for now. I'm not sure this as-is is going to solve the problem.

PR #1957, opened 2026-08-26
Reject an XMLRPC size limit above the SCGI content size limit.

xirvik:
The ceiling on network.xmlrpc.size_limit.set was enforced inside xmlrpc_c.cc, so with the tinyxml2 backend there was no ceiling at all — and the limit it enforced (64 << 20) was unrelated to what actually bounds a request.
 
Move the check to RpcManager::set_size_limit, the funnel both backends pass through, and check against SCgiTask::max_content_size: a limit above that cannot take effect, since the request is refused before the XML is parsed.

### DIFF
--- a/src/rpc/rpc_manager.h
+++ b/src/rpc/rpc_manager.h
@@ -62,7 +62,7 @@ public:
   void                cleanup();
 
   int64_t             size_limit()                  { return m_xmlrpc.size_limit(); };
-  void                set_size_limit(uint64_t size) { m_xmlrpc.set_size_limit(size); };
+  void                set_size_limit(uint64_t size);
 
   int                 dialect()                     { return m_xmlrpc.dialect(); }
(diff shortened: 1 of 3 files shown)


src/rpc/rpc_manager.h
```diff
@@ -62,7 +62,12 @@ class RpcManager {
   void                cleanup();
 
   int64_t             size_limit()                  { return m_xmlrpc.size_limit(); };
-  void                set_size_limit(uint64_t size) { m_xmlrpc.set_size_limit(size); };
+  void                set_size_limit(uint64_t size) {
+    if (size > SCgiTask::max_content_size)
+      throw torrent::input_error("XMLRPC size limit cannot exceed the SCGI content size limit.");
+
+    m_xmlrpc.set_size_limit(size);
+  };
```
rakshasa: We only inline simple one-liners, this should go in .cc

PR #1959, opened 2026-08-26
Keep d.state and the started view in agreement when the core stops a download

### LINKED ISSUE #1958: d.state is left at 0 while the download stays in the started view, so d.start can never restore it
Two code paths write `d.state` directly after the views have been set up, which
leaves the state variable and the `started` view membership disagreeing. Once
they disagree, `d.start` can no longer set the state, however many times it is
issued.

Reproduced against **rtorrent master `c8da93b`** built with **libtorrent master
`4e0aebb`**, in a container, driven over XML-RPC.

## Why the lock is permanent

`d.state` has exactly one writer that can set it back to 1 — the `started`
view's `event_added` (`src/main.cc:266`):

```text
view.event_added = started,{(view.set_not_visible,stopped),(d.state.set,1),(scheduler.simple.added)}
```

`d.start` only asks the view to make the download visible
(`src/command_download.cc:696`):

```cpp
CMD2_FUNC_SINGLE("d.start", "d.hashing_failed.set=0 ;view.set_visible=started");
```

and `View::set_visible()` searches only the non-visible entries
(`src/core/view.cc:182`):

```cpp
iterator itr = std::find_if(begin_filtered(), end_filtered(), entry_is(download));

if (itr == end_filtered())
  return;
```

A download that was never removed from `started` is not among them, so the
function returns, `event_added` does not fire, and `d.state.set=1` is never
reached. The event is edge-triggered, and after a direct write there is no
edge left to trigger.

## The two writers

**1. `DownloadList::hash_done()`, `src/core/download_list.cc:525`** — a download
registered as complete whose hash check finds unfinished chunks:

```cpp
if (rpc::call_command_value("d.complete", rpc::make_target(download)) && !download->is_done()) {
  rpc::call_command("d.state.set", (int64_t)0, rpc::make_target(download));
  download->set_message("Download registered as completed, but hash check returned unfinished chunks.");
}
```

The `resume()` three lines below is already unreachable, because the state was
just set to 0.

**2. `Download::set_root_directory()`, `src/core/download.cc:123`** — refusing
to move an open download whose files are gone:

```cpp
set_message("Cannot change the directory of an open download after the files have been moved.");
rpc::call_command("d.state.set", (int64_t)0, rpc::make_target(this));
control->core()->download_list()->close_directly(this);
```

Both leave the download in the `started` view with the variable at 0.

For contrast, `DownloadFactory` writes the same variable directly and is
correct, because it writes *before* `event.download.inserted_new` places the
download into a view. The defect is the direct write after the views exist, not
the direct write itself. `DownloadList::process_meta_download()` already stops a
download the other way, through `rpc::call_command("d.stop", ...)`.

## Reproduction

A completed, seeding, single-file torrent. `view:started` and `view:stopped` are
read back with `d.multicall2` against those views, so the disagreement is shown
rather than inferred.

**Scenario A — hash check contradicts `d.complete`.** Stop and close, overwrite
the first 4 MiB of the file with random data, then open and start:

```text
0. seeding                         open=1 state=1 active=1 view:started=1 view:stopped=0
1. hash check finished             open=1 state=0 active=0 view:started=1 view:stopped=0
   d.message: 'Download registered as completed, but hash check returned unfinished chunks.'
2.1 d.start                        open=1 state=0 active=0 view:started=1 view:stopped=0
2.2 d.start                        open=1 state=0 active=0 view:started=1 view:stopped=0
2.3 d.start                        open=1 state=0 active=0 view:started=1 view:stopped=0
```

**Scenario B — directory changed after the files moved.** Rename the file away,
call `d.directory.set`, rename it back, then open and start:

```text
0. seeding                         open=1 state=1 active=1 view:started=1 view:stopped=0
   d.directory.set faulted as expected: Cannot change the directory of an open download after the files have been moved.
1. after d.directory.set           open=0 state=0 active=0 view:started=1 view:stopped=0
2.1 d.open + d.start               open=1 state=0 active=0 view:started=1 view:stopped=0
2.2 d.open + d.start               open=1 state=0 active=0 view:started=1 view:stopped=0
2.3 d.open + d.start               open=1 state=0 active=0 view:started=1 view:stopped=0
```

In both, only `d.stop` recovers the download. It makes it visible in `stopped`,
which *is* a change, so that view's `event_added` fires and its
`view.set_not_visible=started` finally takes the download out of `started`. A
later `d.start` can then genuinely add it back. The recovery works by the views
being brought back into agreement, not because stopping means anything to the
user here.

## Consequences

- **The state variable is what survives a restart.** `event.download.inserted_session`
  runs `branch=((d.state)),((view.set_visible,started)),((view.set_visible,stopped))`
  (`src/main.cc:224`), so a download left at `d.state = 0` is placed in `stopped`
  on the next start of the daemon. A download that reached this state while
  seeding comes back stopped and seeds nothing.
- **A download can be active while `d.state` reads 0.** `d.resume` goes straight
  to `DownloadList::resume()` and touches neither the views nor the state
  variable, so it activates a download whose state is locked. ruTorrent's start
  action sends `d.open`, `d.start` and `d.resume` together, which is how this is
  normally met in practice — measured on 0.16.20:

  ```text
  1. hash done, state locked         open=1 state=0 active=0 view:started=1 view:stopped=0
  2. ruTorrent Start (open+start+resume) open=1 state=0 active=1 view:started=1 view:stopped=0
  ```

  Every consumer of `d.state` reports the download as stopped while it transfers.
- **There is no obvious way out for a user.** Start does nothing visible, and
  nothing indicates that stopping first is required.

## Suggested fix

Stop the download through the `d.stop` command in both places instead of writing
the variable. That takes it out of the `started` view, so the state variable is
set through the same event the rest of the code relies on and the two stay in
agreement. It also respects a `d.stop` redefined in `.rtorrent.rc`, which a
direct call to `View::set_not_visible()` would bypass.

A pull request doing this is attached. Both scenarios above pass with it, on the
same torrent and the same session, with only the rtorrent source changed:

```text
1. hash check finished             open=1 state=0 active=0 view:started=0 view:stopped=1
2.1 d.start                        open=1 state=1 active=1 view:started=1 view:stopped=0
```

noctuum:
Fixes #1958.

`hash_done()` and `set_root_directory()` cleared `d.state` while leaving the download in the `started` view. The variable is only set back to 1 by that view's `event_added`, which fires on entry, so a download that never left the view could not be started again: `View::set_visible()` returns early for one already visible, the event does not fire, and `d.start` is a no-op.

Following the review, the stopping now happens in `DownloadList::close_directly()`: it clears `d.state`, adds the download to the `stopped` view through the view rather than a parsed command, and ends with the events it used to swallow — `event.download.paused` when it was active, `hash_removed` and `closed` when it was open. The bookkeeping those share with `pause()` is now `update_paused_state()`, called from both. `hash_done()` does not close the download, so it reaches the same view and variable through the small `set_state_stopped()` the two share.

## Why close_files() exists

`close_directly()` had three callers, and only one of them means "stop this download":

- the refusal in `set_root_directory()`, which does mean it;
- the successful path of `set_root_directory()`, which only needs the files released before they are moved;
- the `d.close.directly` command, which closes files much as `d.close` does, and `d.close` does not clear `d.state` either.

Giving all three the stop would mean that changing a download's directory, or closing its files, also clears the flag that says the user wants it running. That flag is what survives a restart, so the download would come back stopped, and `d.close.directly` would part ways with `d.close`, which leaves it alone. The file-level half is therefore split out as `close_files()`, and the two callers that only need that use it. Their behaviour is unchanged.

## Verification

Two images built from the same context, differing only in the rtorrent source — master `c8da93b` and this branch — both against libtorrent master `4e0aebb`. The same completed torrent and session directory for every run.

    A  corrupt 4 MiB of a completed torrent's file, then open and start it
    B  rename the file away, then call d.directory.set
    C  call d.directory.set on a running download, with its files in place
    D  call d.close.directly on a running download

A and B then issue d.start three times. C and D are regression guards: they must pass before and after.

    check                                                  master   branch
    A  d.state agrees with the started view                  FAIL     PASS
    A  d.start restores d.state                              FAIL     PASS
    B  d.state agrees with the started view                  FAIL     PASS
    B  d.start restores d.state                              FAIL     PASS
    A/B  no other download changed state                     PASS     PASS
    C  a successful directory change does not stop it        PASS     PASS
    D  d.close.directly leaves d.state alone                 PASS     PASS

Recovery was also followed to the end on a live torrent: after the refused hash check the download is stopped, and once started it re-downloads the missing chunks, passes the completion hash check and returns to seeding on its own.

### DIFF
--- a/src/core/download.cc
+++ b/src/core/download.cc
@@ -126,7 +126,7 @@ Download::set_root_directory(const std::string& path) {
     throw torrent::input_error("Cannot change the directory of an open download after the files have been moved.");
   }
 
-  control->core()->download_list()->close_directly(this);
+  control->core()->download_list()->close_files(this);
   file_list->set_root_dir(expand_path(path));
 
   bencode()->get_key("rtorrent").insert_key("directory", path);
--- a/src/core/download.cc
+++ b/src/core/download.cc
@@ -120,7 +120,6 @@ Download::set_root_directory(const std::string& path) {
        !file_stat.update(file_list->front()->frozen_path().str()))) {
 
     set_message("Cannot change the directory of an open download after the files have been moved.");
-    rpc::call_command("d.state.set", (int64_t)0, rpc::make_target(this));
     control->core()->download_list()->close_directly(this);
 
     throw torrent::input_error("Cannot change the directory of an open download after the files have been moved.");
--- a/src/command_download.cc
+++ b/src/command_download.cc
@@ -718,7 +718,7 @@ initialize_command_download() {
   CMD2_DL_V       ("d.pause",      std::bind(&core::DownloadList::pause_default, control->core()->download_list(), std::placeholders::_1));
   CMD2_DL_V       ("d.open",       std::bind(&core::DownloadList::open_throw, control->core()->download_list(), std::placeholders::_1));
   CMD2_DL_V       ("d.close",      std::bind(&core::DownloadList::close_throw, control->core()->download_list(), std::placeholders::_1));
-  CMD2_DL_V       ("d.close.directly", std::bind(&core::DownloadList::close_directly, control->core()->download_list(), std::placeholders::_1));
+  CMD2_DL_V       ("d.close.directly", std::bind(&core::DownloadList::close_files, control->core()->download_list(), std::placeholders::_1));
   CMD2_DL_V       ("d.erase",      std::bind(&core::DownloadList::erase_ptr, control->core()->download_list(), std::placeholders::_1));
   CMD2_DL_V       ("d.check_hash", std::bind(&core::DownloadList::check_hash, control->core()->download_list(), std::placeholders::_1));
 
--- a/src/core/download_list.cc
+++ b/src/core/download_list.cc
@@ -554,7 +554,7 @@ DownloadList::hash_done(Download* download) {
     // If the download was previously completed but the files were
     // f.ex deleted, then we clear the state and complete.
     if (rpc::call_command_value("d.complete", rpc::make_target(download)) && !download->is_done()) {
-      rpc::call_command("d.state.set", (int64_t)0, rpc::make_target(download));
+      set_state_stopped(download);
       download->set_message("Download registered as completed, but hash check returned unfinished chunks.");
     }
 
--- a/src/core/download_list.cc
+++ b/src/core/download_list.cc
@@ -275,7 +275,38 @@ void
 DownloadList::close_directly(Download* download) {
   lt_log_print_info(torrent::LOG_TORRENT_INFO, download->info(), "download_list", "Closing download directly.");
 
+  bool was_active = download->download()->info()->is_active();
+  bool was_open   = download->download()->info()->is_open();
+
   close_files(download);
+  set_state_stopped(download);
+
+  if (was_active) {
+    DL_TRIGGER_EVENT(download, "event.download.paused");
+    update_paused_state(download);
+  }
+
+  if (was_open) {
+    DL_TRIGGER_EVENT(download, "event.download.hash_removed");
+    DL_TRIGGER_EVENT(download, "event.download.closed");
+  }
+}
+
+void
+DownloadList::set_state_stopped(Download* download) {
+  control->view_manager()->find_ptr_throw("stopped")->set_visible(download);
+  rpc::call_command("d.state.set", (int64_t)0, rpc::make_target(download));
+}
+
+void
+DownloadList::update_paused_state(Download* download) {
+  rpc::call_command("d.state_changed.set", torrent::this_thread::cached_seconds().count(), rpc::make_target(download));
+  rpc::call_command("d.state_counter.set", rpc::call_command_value("d.state_counter", rpc::make_target(download)), rpc::make_target(download));
+
+  // If initial seeding is complete, don't try it again when restarting.
+  if (download->is_done() &&
+      rpc::call_command("d.connection_current", torrent::Object(), rpc::make_target(download)).as_string() == "initial_seed")
+    rpc::call_command("d.connection_seed.set", rpc::call_command("d.connection_current", torrent::Object(), rpc::make_target(download)), rpc::make_target(download));
 }
 
 void
@@ -452,15 +483,7 @@ DownloadList::pause(Download* download, int flags) {
     // view.
     DL_TRIGGER_EVENT(download, "event.download.paused");
 
-    auto cached_seconds = torrent::this_thread::cached_seconds().count();
-
-    rpc::call_command("d.state_changed.set", cached_seconds, rpc::make_target(download));
-    rpc::call_command("d.state_counter.set", rpc::call_command_value("d.state_counter", rpc::make_target(download)), rpc::make_target(download));
-
-    // If initial seeding is complete, don't try it again when restarting.
-    if (download->is_done() &&
-        rpc::call_command("d.connection_current", torrent::Object(), rpc::make_target(download)).as_string() == "initial_seed")
-      rpc::call_command("d.connection_seed.set", rpc::call_command("d.connection_current", torrent::Object(), rpc::make_target(download)), rpc::make_target(download));
+    update_paused_state(download);
 
     // Save the state after all the slots, etc have been called so we
     // include the modifications they may make.
--- a/src/core/download_list.cc
+++ b/src/core/download_list.cc
@@ -254,9 +254,11 @@ DownloadList::close(Download* download) {
   }
 }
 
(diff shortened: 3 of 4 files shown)


rakshasa:
These fixes are wrong, setting the state and calling open/close commands have vastly different behaviors.

The bug you encountered needs to be fixed in a different way / place.

noctuum:
@rakshasa Thank you for the review.

Would this direction be reasonable: a single internal function on DownloadList that both clears the state variable and takes the download out of the `started` view, called from hash_done() and set_root_directory() in place of the direct `d.state.set` writes. No client-level commands involved.

The part I am unsure about is the view move. Going through View::set_not_visible() emits event_removed, which pulls in scheduler.simple.removed — pausing the download and resuming others in its place. Avoiding that would mean a quiet variant on View that changes membership without emitting.

I also considered making `d.state` derived from membership of the `started` view rather than a separately stored variable, so the two cannot disagree by construction. That looks expensive: the stored value is what decides the initial view on session load, so it would need a separate persisted field, and the session format changes with it.

Would you want the view events suppressed there, or fired? And if this belongs somewhere else entirely, I will take that instead.

rakshasa:
This is old code, and likely needs to be refactored.

However this seems to be more of an issue of handling the unwind of an error getting thrown.

So cleanly undo it by adding it back to the stopped view in a clean way.

noctuum:
Thanks! The unwind framing is what I was missing.

So: drop the direct `d.state.set` writes in both hash_done() and set_root_directory(), and add the download to the `stopped` view instead, so that view's event_added sets the state. Moving it the way resume() and pause() already move downloads in and out of `active`, rather than through the `d.stop` command.

Will send that unless you see it differently.

rakshasa:
The state still needs to change.

noctuum:
Thanks! So the state write stays, and the view move is added next to it: in hash_done() and set_root_directory(), `view.set_visible=stopped` before the `d.state.set`, so the download ends up in the stopped view and the explicit write stays authoritative regardless of how the view events are configured.

src/core/download.cc
```diff
@@ -120,8 +120,17 @@ Download::set_root_directory(const std::string& path) {
...
+    // The view has to change along with the state variable: a download left in
+    // the 'started' view with the variable at 0 can never be started again, as
+    // the event that restores it fires only on entry to the view.
+    //
+    // Closing first keeps the stop silent, since close_directly() uses
+    // stop_skip_tracker and the pause behind the view change then finds the
+    // download inactive.
     control->core()->download_list()->close_directly(this);
+    rpc::parse_command_single(rpc::make_target(this), "view.set_visible=stopped");
+    rpc::call_command("d.state.set", (int64_t)0, rpc::make_target(this));
```
rakshasa: Reviewing the code, I think this needs to be fixed in `DownloadList::close_directly()`.

And `d.state.set` also moved there, and don't use parse_command for view.

rakshasa:
```
# if active
    DL_TRIGGER_EVENT(download, "event.download.paused");

# if open
  DL_TRIGGER_EVENT(download, "event.download.hash_removed");
  DL_TRIGGER_EVENT(download, "event.download.closed");
```

The above events should be done at the end of the function.

```
    auto cached_seconds = torrent::this_thread::cached_seconds().count();

    rpc::call_command("d.state_changed.set", cached_seconds, rpc::make_target(download));
    rpc::call_command("d.state_counter.set", rpc::call_command_value("d.state_counter", rpc::make_target(download)), rpc::make_target(download));

    // If initial seeding is complete, don't try it again when restarting.
    if (download->is_done() &&
        rpc::call_command("d.connection_current", torrent::Object(), rpc::make_target(download)).as_string() == "initial_seed")
      rpc::call_command("d.connection_seed.set", rpc::call_command("d.connection_current", torrent::Object(), rpc::make_target(download)), rpc::make_target(download));
```

Add a helper function to pause where you put the shared code.

No need to include hashing as that is already guaranteed to have been completed.

noctuum: Done as reviewed: `close_directly()` clears `d.state`, adds the download to the `stopped` view through the view rather than a parsed command, and ends with the paused / hash_removed / closed events. What it shares with `pause()` is now `update_paused_state()`.

One deviation worth your eye: `close_directly()` had three callers and only the refusal in `set_root_directory()` means a stop. The successful directory change and `d.close.directly` only need the files released - clearing the state there would make a move, or a close, come back stopped after a restart, and would part `d.close.directly` from `d.close`. So the file half is split out as `close_files()` and those two use it.

rakshasa:
Looks good.

PR #1972, opened 2026-09-10
Fix false "Session data is invalid" error when loading downloads with empty URI

jakobbg:
### Problem

In recent releases (since commit 324811d — *"Do not fail startup on unreadable session files"*), dynamically added torrents loaded via raw bencode data (such as XMLRPC `load.raw_start` / `load_raw_start` used by ruTorrent and other frontends) are erroneously flagged as failed with:

```
Session data is invalid, ignoring it
```

#### Why this happens:
1. When adding a torrent via raw bencode data, `m_uri` is empty (`""`).
2. In `DownloadFactory::receive_success()`, `download_factory_load_stream((expand_path(m_uri) + ".rtorrent").c_str(), &session_invalid)` was called unconditionally.
3. When `m_uri` is empty, `(expand_path(m_uri) + ".rtorrent")` evaluates to `".rtorrent"` relative to rTorrent's current working directory.
4. If rTorrent was launched from a directory containing a `.rtorrent` subdirectory (most commonly the user's `$HOME` directory where `~/.rtorrent` is the standard rTorrent configuration/session folder), `std::fstream` opens the directory successfully (`stream.is_open()` is true).
5. Reading from an open directory fails with `EISDIR`, causing `stream.good()` to be false.
6. Commit 324811d treats any read failure on an open stream as `*is_invalid = true`, setting `session_invalid = true`.
7. Consequently, `download->set_hash_failed(true)` and `download->set_message("Session data is invalid, ignoring it")` are called, which prevents the torrent from starting or hashing in `DownloadList::resume()`.

Furthermore, session files (`.rtorrent` and `.libtorrent_resume`) only exist and are only relevant when loading saved session state at startup (`m_session == true`). Non-session downloads discard these structures anyway.

### Solution

1. In `DownloadFactory::receive_success()`, only attempt to load session files when `m_session && !m_uri.empty()`.
2. In `download_factory_load_stream()`, check `S_ISREG(sb.st_mode)` via `stat()` to ensure directories and non-regular files are never opened as session files.


src/core/download_factory.cc
```diff
@@ -167,11 +173,16 @@ void
...
   bool session_invalid = false;
 
-  auto rtorrent_object          = download_factory_load_stream((expand_path(m_uri) + ".rtorrent").c_str(), &session_invalid);
-  auto libtorrent_resume_object = download_factory_load_stream((expand_path(m_uri) + ".libtorrent_resume").c_str(), &session_invalid);
+  std::unique_ptr<torrent::Object> rtorrent_object;
+  std::unique_ptr<torrent::Object> libtorrent_resume_object;
 
-  if (session_invalid)
-    lt_log_print(torrent::LOG_ERROR, "%s: %s", session_invalid_message, m_uri.c_str());
+  if (m_session && !m_uri.empty()) {
```
rakshasa:
```
if (m_session) {
  if (m_uri.empty())
     throw input_error
```

jakobbg: Thanks, will add this.

jakobbg (commented):
Updated to throw input_error if m_session is set and m_uri is empty (bb7921b).

PR #1991, opened 2026-09-16
Take ownership of the magnet metadata object

noctuum:
`process_meta_download` allocates the bencode map with `new` and hands it to `try_create_download_from_meta_download`, which serialises it and never deletes it. A `std::unique_ptr` parameter moves the ownership and drops the parse failure `delete`.

A seeding instance and one loading that hash as a magnet, under LSan:

```
master    Direct leak of 48 byte(s) in 1 object(s)
            #1 core::DownloadList::process_meta_download(...) core/download_list.cc:729
          52389 byte(s) leaked in 220 allocation(s)

patched   51135 byte(s) leaked in 205 allocation(s), no frame in that path
```

What leaks is the whole parsed `info` dictionary. `make check` passes.


src/core/manager.cc
```diff
@@ -220,6 +221,8 @@ Manager::try_create_download(const std::string& uri, int flags, const command_li
 
 void
 Manager::try_create_download_from_meta_download(torrent::Object* bencode, const std::string& metafile) {
+  utils::scope_guard guard([bencode]() { delete bencode; });
+
```
rakshasa: Use `std::unique_ptr` instead when passing `bencode`.

noctuum: Done. The parameter is a `std::unique_ptr<torrent::Object>` now and `process_meta_download` moves into it.

PR #2002, opened 2026-09-19
Fix use-after-free when p.multicall commands disconnect peers

xirvik:
p.disconnect erases the peer mid-iteration, freeing it before the row's next command


src/command_download.cc
```diff
@@ -399,13 +401,22 @@ p_multicall(core::Download* download, const torrent::Object::list_type& args) {
...
+  // list and deletes the connection, so iterate over a copy and re-check that the peer is
+  // still connected before each call.
+  const std::vector<torrent::Peer*> peers(connections->begin(), connections->end());
+
+  for (const auto& connection : peers) {
     torrent::Object::list_type& row = result.insert(result.end(), torrent::Object::create_list())->as_list();
 
     for (auto cItr = ++args.begin(); cItr != args.end(); cItr++) {
+      if (std::find(connections->begin(), connections->end(), connection) == connections->end())
+        break;
```
rakshasa, line 418: This is way too inefficient, and disconnecting peers causes reordering of the connection list. (last element gets put in the position of the erased element)

As such commands are not supposed to do anything that causes disconnections during the p.multicall.

A better solution here would be to add a change counter to connection list, and then fail `p.multicall` if there's a change.

xirvik: > A better solution here would be to add a change counter to connection list, and then fail `p.multicall` if there's a change.

Done. (now needs libtorrent PR 972 though).

PR #2004, opened 2026-09-19
Reject out-of-range RPC values instead of narrowing them silently.

xirvik:
Guards choke group max, DHT port, strtoll ERANGE and the 40-char info-hash length.


src/rpc/parse.cc
```diff
@@ -114,8 +115,13 @@ parse_value_nothrow(const char* src, int64_t* value, int base, int unit) {
     throw torrent::input_error("Command::string_to_value_unit(...) received unit <= 0.");
 
   char* last;
+
+  errno  = 0;
   *value = strtoll(src, &last, base);
 
+  if (errno == ERANGE)
+    return src;
```
rakshasa, line 123: Replace with `std::from_chars`.

src/command_tracker.cc
```diff
@@ -142,7 +143,12 @@ initialize_command_tracker() {
       lt_log_print(torrent::LOG_DHT_ERROR, "dht.port.set is no longer supported, use dht.override_port.set", 0);
     });
   CMD2_ANY            ("dht.override_port",     [](auto, auto)        { return torrent::runtime::network_config()->override_dht_port(); });
-  CMD2_ANY_VALUE_V    ("dht.override_port.set", [](auto, auto& value) { return torrent::runtime::network_manager()->set_dht_port(value); });
+  CMD2_ANY_VALUE_V    ("dht.override_port.set", [](auto, auto& value) {
+      if (value < 0 || value > std::numeric_limits<uint16_t>::max())
+        throw torrent::input_error("DHT port must be between 0 and 65535.");
+
+      return torrent::runtime::network_manager()->set_dht_port(value);
```
rakshasa, line 150: This should be a helper function and used for all set port commands.

xirvik: > This should be a helper function and used for all set port commands.

Done. Well, except the string-parsing commands network.listen.port.range.set, network.scgi.open_port/open_local, dht.add_node, d.add_peer)

src/rpc/parse.cc
```diff
@@ -108,15 +110,50 @@ value_fits_shifted(int64_t value, int shift) {
...
+    first++;
+
+  if ((base == 0 || base == 16) && first[0] == '0' && (first[1] == 'x' || first[1] == 'X') &&
+      std::isxdigit(static_cast<unsigned char>(first[2]))) {
+    first += 2;
+    base = 16;
+
+  } else if (base == 0) {
+    base = *first == '0' ? 8 : 10;
+  }
```
rakshasa, line 130: Why are you parsing -/+ instead of leaving that to std::from_char?

We don't use negative numbers when calling with base set to 16, and we don't use base 8.

So just a simple check for `0x` if base==16, and let std::from_chars() handle all other cases.

src/rpc/parse.cc
```diff
@@ -125,6 +162,11 @@ parse_value_nothrow(const char* src, int64_t* value, int base, int unit) {
     return src;
   }
 
+  if (!value_from_magnitude(magnitude, negative, value))
+    return src;
```
rakshasa: This is wrong, unit is only indicating the default unit, the switch below decided the unit to actually use.

src/rpc/parse.cc
xirvik: Done, note that from_chars doesn't parse for example +5.

src/rpc/parse.cc
```diff
@@ -113,10 +115,28 @@ parse_value_nothrow(const char* src, int64_t* value, int base, int unit) {
...
-  *value = strtoll(src, &last, base);
+  const char* first = src;
+
+  while (std::isspace(static_cast<unsigned char>(*first)))
+    first++;
+
+  if ((base == 0 || base == 16) && first[0] == '0' && (first[1] == 'x' || first[1] == 'X') &&
+      std::isxdigit(static_cast<unsigned char>(first[2]))) {
+    first += 2;
+    base = 16;
```
rakshasa, line 126:
```
if (base == 16 && first[0] == '0' && (first[1] == 'x' || first[1] == 'X')) {
    first += 2;
}
```

https://cppreference.com/cpp/utility/from_chars

xirvik: Dropped the isxdigit guard, but we need to keep base == 0 since nothing passes an explicit base, so 0x... would return src unchanged.

rakshasa: `"0x" or "0X" prefixes are not recognized if base is 16`

Base 0 will detect the prefix.

PR #2008, opened 2026-09-19
Bound JSON-RPC input by size and nesting depth

xirvik:
Scan nesting before parsing and apply network.xmlrpc.size_limit to the JSON path too.


src/rpc/jsonrpc.cc
```diff
@@ -223,11 +223,71 @@ handle_notification(const json& request) noexcept {
...
+    auto err_str = json_error(JSONRPC_INVALID_REQUEST_ERROR, "content size exceeds maximum RPC limit", nullptr).dump();
+
+    return callback(err_str.c_str(), err_str.size());
+  }
+
+  if (!json_within_depth_limit(in_buffer, length)) {
+    auto err_str = json_error(JSONRPC_INVALID_REQUEST_ERROR, "maximum nesting depth exceeded", nullptr).dump();
+
+    return callback(err_str.c_str(), err_str.size());
+  }
```
rakshasa: This is too expensive, we shouldn't be parsing json twice just for a depth check.

xirvik: Fixed

PR #2010, opened 2026-09-19
Write session files with O_NOFOLLOW and mode 0600

xirvik:
Also make directory_entry::is_file() report the real type so symlinks are skipped.


src/session/download_storer.cc
```diff
@@ -116,28 +117,36 @@ is_correct_format(const std::string& f) {
...
 
-  output << stream.rdbuf();
+  if (fd < 0)
+    throw torrent::storage_error("failed to open file for writing : " + path);
 
-  if (!output.good())
-    throw torrent::storage_error("failed to write stream to file : " + path);
+  const auto  data      = stream.str();
+  std::size_t remaining = data.size();
+  const char* cursor    = data.data();
```
rakshasa, line 134: https://en.cppreference.com/cpp/io/basic_stringstream/view

PR #2015, opened 2026-09-19
Guard download pointers against event handlers that erase the download

xirvik:
Re-find the download after each event, matching the existing check in confirm_finished.


src/core/download_list.cc
```diff
@@ -196,14 +196,27 @@ DownloadList::erase(iterator itr) {
...
   lt_log_print_info(torrent::LOG_TORRENT_INFO, (*itr)->info(), "download_list", "Erasing download.");
 
+  torrent::HashString infohash = (*itr)->info()->hash();
+
   // Makes sure close doesn't restart hashing of this download.
   (*itr)->set_hash_failed(true);
 
   close(*itr);
+
+  itr = find(infohash);
```
rakshasa, line 206: This actually introduces possible collisions if the same infohash is loaded again.

Better to use e.g. std::shared_ptr and std::weak_ptr to a lifetime handle and check it. Also possibly guard against erase while in erase.

PR #2016, opened 2026-09-19
Make SessionManager::m_active atomic

xirvik:
It is written under m_mutex but read under m_pending_builds_mutex in one place.


src/session/session_manager.h
```diff
@@ -101,7 +102,7 @@ class SessionManager {
 
   align_cacheline std::mutex   m_mutex;
 
-  bool                         m_active{};
+  std::atomic<bool>            m_active{};
```
rakshasa, line 105:
```
align_cacheline  

std::atomic<bool>  m_active{};

align_cacheline  

std::mutex   m_mutex;
```

xirvik: > ```
> align_cacheline  
> 
> std::atomic<bool>  m_active{};
> 
> align_cacheline  
> 
> std::mutex   m_mutex;
> ```

Done

rakshasa:
The align only guarantees you start at a new cacheline, the following variables share that cacheline so you can't mix them up.

PR #2018, opened 2026-09-19
Serialize session saves for each download path

### LINKED ISSUE #2017: Concurrent session saves can leave a new torrent out of the restored session
Overlapping saves for one download can leave its main `.torrent` session file uncommitted. The download remains visible until rTorrent restarts, then disappears from the session. Its downloaded data is still on disk.

Reproduced with rTorrent/libtorrent 0.16.23 on Linux, including a clean SIGINT shutdown with exit status 0. The session directory contains `.torrent.new`, `.torrent.rtorrent` and `.torrent.libtorrent_resume`, but no committed `.torrent` file.

Related to #1867. This case uses the stock insertion events: the full save starts and writes its temporary files successfully before another save interferes with it.

## What the trace shows

Two `SessionManager` workers write the same temporary paths. The first is saving the full torrent; the second is saving resume data.

```text
worker A: write HASH.torrent.new
worker A: write and close HASH.torrent.rtorrent.new
worker B: write and sync HASH.torrent.rtorrent.new
worker B: rename HASH.torrent.rtorrent.new -> HASH.torrent.rtorrent = 0
worker A: open HASH.torrent.rtorrent.new, O_WRONLY = -1 ENOENT
```

rTorrent reports:

```text
Storage errors saving session data for download: ignored:0 : failed to open file descriptor for fsync
```

Worker A exits before committing `HASH.torrent.new`. Later resume saves and a clean shutdown don't recreate that main file, so the next startup ignores the unfinished session entry.

## Reproduction

1. Start rTorrent with an empty session directory and its local SCGI socket enabled.
2. Load a small synthetic torrent with `load.normal`, then call `session.save` while its initial full save is still running.
3. Let the saves finish, stop rTorrent with SIGINT and restart with the same session directory.

This first occurred without instrumentation on a busy disk. To control the timing, I repeated it under strace with `-f -e inject=openat:delay_enter=500ms:when=6`. In the full-save worker, the sixth `openat` reopens `.torrent.rtorrent.new` for synchronization. The delay lets the resume worker rename that file first. No error result was injected; the delayed open returned ENOENT from the filesystem.

With the overlap forced, 0/5 torrents returned after restart. Waiting for the initial full save before requesting another save preserved 5/5 under the same delay. The single-torrent case also failed on tmpfs. All payloads stayed unchanged.

The built-in `session_save` timer reproduced it too, with its first tick moved forward in the test to overlap a delayed initial save. It doesn't require a custom insertion hook.

## Proposed fix

`process_next_save_request_unsafe()` currently takes the next pending request without checking whether a worker is already saving that path. It should leave that request queued until the active save finishes, while allowing requests for other paths to proceed. Shutdown's `flush_all_and_wait_unsafe()` needs the same exclusion and must wait when every remaining request conflicts with an active save.

This keeps the existing pending-request coalescing and parallel saves across different torrents. The same scheduling code is present in 0.16.22, although the runtime tests above used 0.16.23.

Rid:
Fixes #2017.

An initial full save and a resume save can write the same `.new` files at the same time. If the resume worker renames a sidecar before the full-save worker reopens it, the full save fails with ENOENT and never commits the main torrent file. The torrent then disappears after restart.

Leave pending requests queued while their path has an active writer. Select the next request whose path is free so other torrents can keep saving. Apply the same selection during shutdown, waiting for a worker to finish when all remaining paths are busy.

Validation on Debian 13 with GCC 14 and libtorrent `0d7118939a427b35739d8b44b54b567703aebb0a`:

- All four regression cases failed on unpatched rTorrent `1b15b7c518dc3a46270e4451248e6a5dfa83078d` and passed with this fix. They cover overlapping initial and resume saves, queued updates, progress for another torrent, and shutdown with a save active. The script controls the interleaving by delaying an open; it does not inject a syscall error.
- The patched runs verified clean exits, committed session files, restored hashes and custom fields after restart, and unchanged payloads.
- `make -j4 check` passed both test executables with no skips.

Those checks ran before the test files were removed from this PR. The two production source files are unchanged from the tested version. This PR contains only the session-manager fix.


rakshasa:
Please don't add CI tests to the main repository.

https://github.com/rakshasa/rtorrent-docker is the only CI I'm currently using.

Rid:
Thanks @rakshasa

I've removed the tests and force-pushed the branch. Would you like me to add the regression test in a separate PR to rtorrent-docker?

rakshasa:
I haven't gotten around to properly reviewed your PR and it touches parts of the code I'm not going to change without careful consideration.

BTW if anyone can get rtorrent-docker running and provide fixes to it I'd be glad, as it seem I'm the only one who currently runs it.

PR #2019, opened 2026-09-19
Match multicall struct members by name instead of by position

xirvik:
XML-RPC structs are unordered, so params before methodName faulted the batch.


rakshasa:
Is this really an issue? Looks like extra code for something we might as well reject.

xirvik:
> Is this really an issue? Looks like extra code for something we might as well reject.

Agreed. That what it does now.

